PR #8, opened 2022-04-25
Add ansible example.

wangtt03:
Add an example to use Ansible script to deploy DeepGNN package across servers.


mortid0:
@wangtt03 could you update PR with a new CI?

wangtt03:
Updated the ansible test cases using new CI. Thanks!

examples/distributed/README.md
```diff
@@ -0,0 +1,63 @@
+# Launch distributed DeepGNN in cluster
+
```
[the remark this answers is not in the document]
coledie: Maybe we should also add a note like these scripts are only useful for >10gb datasets or something so new users know then can just focus on pytorch/tf examples.

examples/distributed/README.md
```diff
@@ -0,0 +1,63 @@
...
+      hosts:
+        azure-linux-vm01: # machine alias
+          ansible_host: x.x.x.x # machine ip
+      vars:
+        ansible_connection: paramiko
+        ansible_user: azureuser # ssh user
+        ansible_ssh_pass: "password" # ssh password
+    ```
+
+2. Windows nodes
```
mortid0, line 24: where is a CI test for windows?

examples/distributed/README.md
```diff
@@ -0,0 +1,63 @@
...
+        ansible_password: "password"
+        ansible_python_interpreter: python
+    ```
+
+### Usage
+
+Go to `distributed/ansible` folder and invoke following command:
+
+```Shell
+python distributed.py \
```
mortid0, line 49: there is a typo here, `distribute.py` instead of `distributed.py`.
How do we make sure this document is in sync with the python file?

examples/distributed/README.md
```diff
@@ -0,0 +1,63 @@
+# Launch distributed DeepGNN in cluster
+
+## Using ansible
+
```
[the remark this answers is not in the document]
wangtt03: Added some comments about the usage of each ansible configuration files. Thanks!

.gitignore
```diff
@@ -22,5 +22,5 @@ build/*
 #IDE related
 .vscode
 
-# compiler toolchain
-.dt11*
+# ansible artifacts
```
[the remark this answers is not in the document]
wangtt03: A mistake when merging the code, fixed. Thanks!

examples/distributed/README.md
wangtt03: I did some investigation about how to do that in github action because ansible doesn't support windows control node. If we want to test it on windows, we need to setup 2 agents, one for linux to run the ansible script and 1 for windows to run the GE. I didn't find a good solution yet because it is a little bit complicated to orchestrate these 2 nodes in github action. We need to make sure these two agents are in same vnet or a public dns which is used to find each other...

Do you have some suggestions? Thanks!

examples/distributed/README.md
wangtt03: Good catch! Thanks!

Currently I copied the commands in `run.sh` to README.md to make sure they are the same (at least these commands are runnable in run.sh). At this time let me do it manually. Maybe next step we can do it using notebook. Does it make sense?

examples/distributed/ansible/playbooks/ubuntu.yml
```diff
@@ -0,0 +1,43 @@
...
+  become: true
+  package:
+    name:
+      - software-properties-common
+    state: present
+
+- name: Add deadsnake python repo
+  when: pythonVersion["stdout"] is version('3.6', '<=')
+  ansible.builtin.apt_repository:
+    repo: ppa:deadsnakes/ppa
```
coledie, line 28: Possible to avoid PPA? If ubuntu >= 20 I think python 3.8 is the default.

examples/distributed/ansible/playbooks/start.yml.template
```diff
@@ -0,0 +1,25 @@
...
+  {% endif %}
+
+  {% if item[1] == "windows" %}
+
+- hosts: {{ item[0] }}
+  tasks:
+  - name: Start graph engine
+    win_shell: start /b python -m deepgnn.graph_engine.snark.server --data_dir {{ data_dir }} --partitions {{ item[2] }} --port {{ port }}
+    args:
+      executable: cmd.exe
```
coledie, line 22: Can this if else just be placed around the shell? I think it is more readable like that.

coledie (approved):

coledie:
Is it safe to close this? We can always re-open later. @wangtt03 @mortid0

PR #11, opened 2022-04-26
Enhance setup.py

mortid0:
* Add classifiers to the project, to make it easy to find and categorize.
* Drop requirement <4, since Python-Requires doesn’t support upper bounds 
* Add a long description to torch/tf packages to point to deepgnn-ge packages for details

### DIFF
--- a/src/python/setup.py
+++ b/src/python/setup.py
@@ -65,9 +65,23 @@ def graph_engine(version: str):
             "Source": "https://github.com/microsoft/DeepGNN",
         },
         include_package_data=True,
-        python_requires=">=3.7, <4",
+        python_requires=">=3.7",
         cmdclass={"clean": clean},
         license="MIT",
+        classifiers=[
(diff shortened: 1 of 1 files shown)


src/python/setup.py
```diff
@@ -93,16 +107,31 @@ def deepgnn_tf(version: str):
         name="deepgnn-tf",
         version=version,
         description="DeepGNN algorithms for tensorflow.",
+        long_description="See [DeepGNN package](https://pypi.org/project/deepgnn-ge/) for detailed description.",
```
coledie, line 110: Many repos make this == the README with file = open(filename) and file.readlines() if you are interested.

coledie (approved):

src/python/setup.py
mortid0: That's what we do on line 40. Here we reference the main package.

PR #13, opened 2022-04-28
add the set logger

DoWhatILove:


DoWhatILove:
@coledie as discussed, would you please take a look?

coledie (approved):

src/python/deepgnn/logging_utils.py
```diff
@@ -126,6 +126,22 @@ def setup_default_logging_config(enable_telemetry: bool = False):
     _init = True
 
 
+def set_logger(customized_logger: logging.Logger) -> None:
```
mortid0, line 129: @DoWhatILove Could you elaborate why do we need a setlogger function instead of adding custom handlers to the existing logger? This seems to be a standard practice:
https://docs.python.org/3/howto/logging-cookbook.html#using-logging-in-multiple-modules
Standard library also doesn't have any setters:
https://docs.python.org/3/library/logging.html#module-level-functions

DoWhatILove:
@coledie, it seems there is something wrong for setup.py.

mortid0:
> @coledie, it seems there is something wrong for setup.py.

Sorry about that @DoWhatILove, could you rebase?
CI should be green now: https://github.com/microsoft/DeepGNN/runs/6219215059?check_suite_focus=true

DoWhatILove:
@mortid0 , it  seems that I made the wrong rebase, as there were 20 files changed which contains the active PRs.

this is what I do: update my forked repo from this DeepGNN repo, then rebase "git rebase main", finally "git push". do you know what was wrong?

mortid0:
> @mortid0 , it seems that I made the wrong rebase, as there were 20 files changed which contains the active PRs.
> 
> this is what I did: updated my forked repo from this DeepGNN repo, then rebase "git rebase main", finally "git push". do you know what was wrong?

I think your forked repo `main` is different from `microsoft/DeepGNN` `main` and rebase happened on the forked main, rather than here. You'll need to reorder commits: 
image

PR #14, opened 2022-04-28
Add CI

mortid0:
Add a CI workflow that is triggered on PRs targeting main branch.
Every CI step is implemented as a [GitHub action](https://docs.github.com/en/actions/creating-actions/about-custom-actions), so it can be shared in different workflows. Bazel artifacts produced in different actions are also shared inside a single job. Example of run triggered by a PR: https://github.com/mortid0/DeepGNN/runs/6214106146?check_suite_focus=true

Every action has a separate workflow to help with testing individual component in branches: lint, tests, examples and benchmarks. These actions can be launched from the Actions tab:
image


.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+      - main
+  pull_request_target:
+    branches:
+      - main
+
+jobs:
+  pre-commit:
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9"]
```
coledie, line 15: pre-commit runs with 3.8 and 3.9 - tests run with 3.7 and 3.8.

PR #17, opened 2022-04-29
make the logger customized with set logger function

DoWhatILove:
make the logger customized with set logger function

we have two options to make customized logger:
1. client makes changes to LOGGING
2. create new set_logger function which allows client to pass customized logger

in this PR, we are implementing the second one.

### DIFF
--- a/src/python/deepgnn/__init__.py
+++ b/src/python/deepgnn/__init__.py
@@ -34,5 +34,6 @@ from .logging_utils import (
     get_current_user,
     log_telemetry,
     get_logger,
+    set_logger,
     setup_default_logging_config,
 )
--- a/src/python/deepgnn/logging_utils.py
+++ b/src/python/deepgnn/logging_utils.py
@@ -127,7 +127,7 @@ def setup_default_logging_config(enable_telemetry: bool = False):
 
 
 def set_logger(customized_logger: logging.Logger) -> None:
-    ''' client could provide the contomized logger.
+    ''' client could provide the customized logger.
 
     This function will override the underlying logger `_logger`.
     '''
--- a/src/python/deepgnn/logging_utils.py
+++ b/src/python/deepgnn/logging_utils.py
@@ -126,6 +126,22 @@ def setup_default_logging_config(enable_telemetry: bool = False):
     _init = True
 
 
+def set_logger(customized_logger: logging.Logger) -> None:
+    ''' client could provide the contomized logger.
+
+    This function will override the underlying logger `_logger`.
+    '''
+    global _logger
+
+    _logger_lock.acquire()
+
+    try:
+        _logger = customized_logger
+
+    finally:
+        _logger_lock.release()
+
+
 def get_logger():
     global _logger
     global _init


DoWhatILove:
@coledie  and @mortid0 , please review it. I recreated the pull request.

DoWhatILove:
> > @coledie and @mortid0 , please review it. I recreated the pull request.
> 
> 1. @DoWhatILove, please fix linter issues: https://github.com/microsoft/DeepGNN/runs/6223151209?check_suite_focus=true
>    Status checks are not yet fully enabled, because we want to check how stable they are first.

I have checked in a commit which format the python via black. however it seems doesn't trigger the CI to rerun the lint check.

mortid0:
> @mortid0 , I have replied in the previous PR. but not sure whether you could see it or not.
> 
> image

1. No one can see reply, because it is in the pending state.
image

Comments need to be posted.

2. The question I am asking is *why* did you pick `set_logger` over handlers, not *what* are the options. [Logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html#using-logging-in-multiple-modules) recipe is to use handlers.

DoWhatILove:
the reason of using __set_logger__ because in our experiments, we have our owner get_logger which creates the logger used for different AML components. For example, the command component is using the ConsoleHandler, and the parallel component is using the FileHandler. We would like to pass our logger into deepgnn package, so that deepgnn could do the right logging for different AML components.

Besides that, we also need to change the format string by prefixing "SystemLog:".

If we make the change based on the logging recipes, then we need to create the new function such as add_handler, and assign_formatter. Please let me know if I misunderstand anything.

mortid0:
> the reason of using **set_logger** because in our experiments, we have our owner get_logger which creates the logger used for different AML components. For example, the command component is using the ConsoleHandler, and the parallel component is using the FileHandler. We would like to pass our logger into deepgnn package, so that deepgnn could do the right logging for different AML components.
> 
> Besides that, we also need to change the format string by prefixing "SystemLog:".
> 
> If we make the change based on the logging recipes, then we need to create the new function such as add_handler, and assign_formatter. Please let me know if I misunderstand anything.
> 
> Or as mentioned in the last reply, we need to change the LOGGING by specifying the format string and handler.

Passing loggers around is a specific pattern to avoid: https://docs.python.org/3/howto/logging-cookbook.html#using-loggers-as-attributes-in-a-class-or-passing-them-as-parameters

There is no need to do this configuration in a library, but you can [configure existing logger](https://docs.python.org/3/howto/logging-cookbook.html#adding-handlers-other-than-nullhandler-to-a-logger-in-a-library) to output messages directly in `main.py`: 
```python
import logging
from deepgnn.log_consts import LOG_NAME_DEEPGNN

dg_logger = logging.getLogger(LOG_NAME_DEEPGNN)
fh = logging.FileHandler('training.log')
formatter = logging.Formatter('SystemLog: %(asctime)s - %(name)s - %(levelname)s - %(message)s')
fh.setFormatter(formatter)
dg_logger.addHandler(fh)


```

DoWhatILove:
> > the reason of using **set_logger** because in our experiments, we have our owner get_logger which creates the logger used for different AML components. For example, the command component is using the ConsoleHandler, and the parallel component is using the FileHandler. We would like to pass our logger into deepgnn package, so that deepgnn could do the right logging for different AML components.
> > Besides that, we also need to change the format string by prefixing "SystemLog:".
> > If we make the change based on the logging recipes, then we need to create the new function such as add_handler, and assign_formatter. Please let me know if I misunderstand anything.
> > Or as mentioned in the last reply, we need to change the LOGGING by specifying the format string and handler.
> 
> Passing loggers around is a specific pattern to avoid: https://docs.python.org/3/howto/logging-cookbook.html#using-loggers-as-attributes-in-a-class-or-passing-them-as-parameters
> 
> There is no need to do this configuration in a library, but you can [configure existing logger](https://docs.python.org/3/howto/logging-cookbook.html#adding-handlers-other-than-nullhandler-to-a-logger-in-a-library) to output messages directly in `main.py`:
> 
> ```python
> import logging
> from deepgnn.log_consts import LOG_NAME_DEEPGNN
> 
> dg_logger = logging.getLogger(LOG_NAME_DEEPGNN)
> fh = logging.FileHandler('training.log')
> formatter = logging.Formatter('SystemLog: %(asctime)s - %(name)s - %(levelname)s - %(message)s')
> fh.setFormatter(formatter)
> dg_logger.addHandler(fh)
> ```

@mortid0 thanks for the example.

Let me point out one thing: in the `logging_utils` of __deepgnn__, it will override the customized config because of `LOGGING` in the function:
```python
def setup_default_logging_config(enable_telemetry: bool = False):
    global _init
    if logging.getLoggerClass() is logging.Logger:
        ## initialize logging config for default logger
        AzureAppInsightFilter.ENABLE_TELEMETRY = enable_telemetry
        logging.config.dictConfig(LOGGING)
        try:
            add_azure_handler(LOG_NAME_DEEPGNN)
        except:
            AzureAppInsightFilter.ENABLE_TELEMETRY = False
    _init = True
```

the LOGGING is defined as:
```python
LOGGING = {
    "version": 1,
    "formatters": {
        "detailed": {
            "format": "[%(asctime)s] {%(filename)s:%(lineno)d} %(levelname)s - %(message)s"
        }
    },
    "handlers": {
        "console": {
            "class": "logging.StreamHandler",
            "formatter": "detailed",
            "filters": [],
        }
    },
    "loggers": {
        LOG_NAME_DEEPGNN: {
            "level": "INFO",
            "propagate": False,
            "handlers": ["console"],
        },
        "tensorflow": {"level": "INFO", "propagate": False, "handlers": ["console"]},
    },
}
```

if we follow the suggested practice in the logging recipe, then I have to modify this `setup_default_logging_config`, does it make sense to you?

DoWhatILove:
> 

after thinking twice, actually applications using `deepgnn` could call `get_logger` of `deepgnn` firstly to initialize the `_logger`, then configure the handler and formatter to override default settings, is it ok?

mortid0:
> > 
> 
> after thinking twice, actually applications using `deepgnn` could call `get_logger` of `deepgnn` firstly to initialize the `_logger`, then configure the handler and formatter in the application code.
> 
> That being said, I prefer to change the function `setup_default_logging_config`.

Ah, yes, good catch! Can you use `deepgnn.loggin_utils.get_logger` instead of `logging.getLogger(LOG_NAME_DEEPGNN)`. The code in main module should look like this:
```Python
import logging
import deepgnn.loggin_utils as dg_logger # Changed

dg_logger = dg_logger.get_logger()          # Changed
fh = logging.FileHandler('training.log')
formatter = logging.Formatter('SystemLog: %(asctime)s - %(name)s - %(levelname)s - %(message)s')
fh.setFormatter(formatter)
dg_logger.addHandler(fh)
```

DoWhatILove:
> > > 
> > 
> > 
> > after thinking twice, actually applications using `deepgnn` could call `get_logger` of `deepgnn` firstly to initialize the `_logger`, then configure the handler and formatter in the application code.
> > That being said, I prefer to change the function `setup_default_logging_config`.
> 
> Ah, yes, good catch! Can you use `deepgnn.loggin_utils.get_logger` instead of `logging.getLogger(LOG_NAME_DEEPGNN)`? The code in main module should look like this:
> 
> ```python
> import logging
> import deepgnn.loggin_utils as dg_logger # Changed
> 
> dg_logger = dg_logger.get_logger()          # Changed
> fh = logging.FileHandler('training.log')
> formatter = logging.Formatter('SystemLog: %(asctime)s - %(name)s - %(levelname)s - %(message)s')
> fh.setFormatter(formatter)
> dg_logger.addHandler(fh)
> ```

Then I could close this PR, right? It seems all could be done on the application side.

mortid0:
> > > after thinking twice, actually applications using `deepgnn` could call `get_logger` of `deepgnn` firstly to initialize the `_logger`, then configure the handler and formatter in the application code.
> > > That being said, I prefer to change the function `setup_default_logging_config`.
> > 
> > 
> > Ah, yes, good catch! Can you use `deepgnn.loggin_utils.get_logger` instead of `logging.getLogger(LOG_NAME_DEEPGNN)`? The code in main module should look like this:
> > ```python
> > import logging
> > import deepgnn.loggin_utils as dg_logger # Changed
> > 
> > dg_logger = dg_logger.get_logger()          # Changed
> > fh = logging.FileHandler('training.log')
> > formatter = logging.Formatter('SystemLog: %(asctime)s - %(name)s - %(levelname)s - %(message)s')
> > fh.setFormatter(formatter)
> > dg_logger.addHandler(fh)
> > ```
> 
> Then I could close this PR, right? It seems all could be done on the application side. @mortid0

Yes, I believe so.

DoWhatILove:
the requirement could be fulfilled on the application side.

PR #20, opened 2022-05-02
Converter accept linear input format

coledie:
Externalize graph format
- Make easier to convert from parquet/orvo to our input format.
- Support neighbor inversion // sorting by dst nodes.
- JSON can't handle high number of edges well.

### DIFF
--- a/src/python/deepgnn/graph_engine/snark/BUILD
+++ b/src/python/deepgnn/graph_engine/snark/BUILD
@@ -16,7 +16,7 @@ py_library(
         "convert.py",
         "meta_merger.py",
         "converter/__init__.py",
-        "converter/json_converter.py",
+        "converter/converter.py",
         "converter/options.py",
         "decoders.py",
         "dispatcher.py",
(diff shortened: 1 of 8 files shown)


mortid0 (commented):
1. Could you add a description, what is going to be the format? From the code it looks like you'll need 5 fields, src, dst, type, weight and feature data, with records separated with new lines. How this format can be extended to support binary encoding, rather than text(e.g. avro/protocol buffers)?
2. You should use external sort, rather than an in memory sort. Especially since you are using lists and strings..
3. convert.py looks very hairy tbh. depending on decoder type we decide which writers to use. It might be better to put this logic inside writer class implementation.

coledie:
Initial benchmarks,
JSON vs Linear
1M / 1 edge 170.32 vs 237.94
10k / 20 edges 89.48 vs 151.85

- all time in converter output

mortid0:
> Initial benchmarks, JSON vs Linear 1M / 1 edge 170.32 vs 237.94 10k / 20 edges 89.48 vs 151.85
> 
> * all time in converter output

linear should be way faster than json :) json decoding is actually expensive.

PR #22, opened 2022-05-11
Run examples scripts

mortid0:
With this change we'll finish CI migration with running examples scripts on every PR.
One extra input added to actions is a disk cache. It is added later to bazel commands.
The cache key is based on PR's base branch name (typically origin/main), so bazel will 
reuse artifacts from previous builds wherever possible.


examples/pytorch/requirements.txt
```diff
@@ -1,3 +1,4 @@
-torch-geometric
-torch-scatter
-torch-sparse
+horovod[torch]==0.24.3
```
coledie, line 1: We should be able to remove this file since these requirements are already in the main one?

.github/workflows/ci.yml
```diff
@@ -36,7 +39,129 @@ jobs:
...
+          python-version: "3.8"
+      - name: Upload examples
+        uses: actions/upload-artifact@v3
+        with:
+          name: examples
+          path: examples/*
+      - name: build wheel
+        uses: ./.github/actions/wheel
+        with:
+          package_version: "0.1.1"
```
coledie, line 70: This should not be a fixed value right?

.github/actions/wheel/action.yml
```diff
@@ -0,0 +1,86 @@
+name: "Wheel"
+description: "Build wheels"
+inputs:
+  package_version:
+    description: 'Package version'
+    required: true
+    default: '0.1.1'
```
coledie, line 7: Would it be better to default to some dev version? 0.0.dev5555

coledie (approved):

examples/pytorch/requirements.txt
mortid0: This is needed for E2E testing without bazel: pip install wheels on a clean instance and run scripts. I added horovod and versions here [for caching](https://github.com/actions/setup-python#caching-packages-dependencies).

.github/workflows/ci.yml
mortid0: This is needed for E2E testing, not to release wheels. Release is handled by a separate `wheels` workflow, instead of an action here.

.github/actions/wheel/action.yml
mortid0: Well, only if we add an extra logic to [setup.py](https://github.com/microsoft/DeepGNN/blob/main/src/python/setup.py#L136) file for dev versions to depend on each other(e.g. deepgnn-tf dev version should also depend on deepgnn-ge dev version). We save artifacts with the workflow instead of pushing them to a feed, which means we don't have to be smart :)

PR #27, opened 2022-05-25
Enable feature for different types.

wangtt03:


src/python/deepgnn/graph_engine/snark/converter/json_converter.py
```diff
@@ -455,27 +455,73 @@ def __add_sparse(node, feature, tp, container):
...
             + values_buf
         )
+
+        if int(k) in container:
+            newId = __get_unassigned_index(node)
+            get_logger().warning(
+                f"Duplicate feature ids found for a node, original: {k}, replaced with: {newId}"
+            )
+            k = newId
+
```
coledie, line 465: Why do we not require users to submit the correct/new ids by themselves?

coledie, line 465: Why do we not require users to submit the correct/new ids by themselves?

wangtt03: Sometimes it is easy to make mistake when assigning the feature index, and the converter process will be pending there. This issue is discussed with LinkedIn team in order to make the converter process easier for them.

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -46,7 +60,12 @@ def triangle_graph_json(folder):
             "node_weight": 1,
             "neighbor": {"0": {}, "1": {"5": 1}},
             "uint64_feature": {},
-            "float_feature": {"0": [1], "1": [-0.03, -0.04]},
+            "float_feature": [[1], [-0.03, -0.04]]
+            if formatType == FeatureFormat.Array
+            else {"0": [1], "1": [-0.03, -0.04]},
+            "double_feature": {"0": [1], "1": [-0.03, -0.04]}
```
mortid0, line 66: What is the expected return value when we fetch features with id = 0, if we have 2 features with the same id, but different types, float and double?

mortid0 (commented):
Please fill the description field in this PR: why it is needed and what problem does it solve.

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
wangtt03: For each feature we have only 1 unique id. If there are features with same id, we will assign a non-existing id for it.

src/python/deepgnn/graph_engine/snark/converter/json_converter.py
```diff
@@ -455,27 +455,73 @@ def __add_sparse(node, feature, tp, container):
...
     if feature not in node or node[feature] is None:
         return
     for k in node[feature]:
         values = node[feature][k]
         buf = (tp * len(values))()
         buf[:] = values
-        assert int(k) not in container, "Duplicate feature ids found for a node"
+
+        if int(k) in container:
+            newId = __get_unassigned_index(node)
```
mortid0, line 489: nit: please follow [PEP8](https://peps.python.org/pep-0008/#function-and-variable-names), `new_id` instead of `newId`

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
wangtt03: By assigning a hidden id is just a tradeoff, in case that users make mistake by setting ids. We could suggest users to use array or TSV format to convert the data.

The test cases are covered [here](https://github.com/microsoft/DeepGNN/blob/tiantiaw/update_converter/src/python/deepgnn/graph_engine/snark/tests/convert_test.py#:~:text=param%20%3D%20%5B,%5D).

src/python/deepgnn/graph_engine/snark/converter/json_converter.py
```diff
@@ -455,27 +455,73 @@ def __add_sparse(node, feature, tp, container):
...
+    if feature_name in node and isinstance(node[feature_name], typing.List):
+        index_str = [
+            __get_unassigned_index(node) for _ in range(len(node[feature_name]))
+        ]
+        node[feature_name] = dict(zip(index_str, node[feature_name]))
+
+
 def convert_features(node: typing.Any):
     """Convert the node's feature into bytes sorted by index."""
+    if "feature_indexes" not in node:
```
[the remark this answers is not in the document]
wangtt03: Tested in local box using Reddit dataset and converted into 1 partition, the performance is much similar, original map based  converter takes: 0:15:02.232133 and array based converter takes: 0:14:45.745795.

src/python/deepgnn/graph_engine/snark/converter/json_converter.py
wangtt03: Done, thanks!

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
wangtt03: > A new id is hidden in a conversion log? It will be hard to find it for a graph with billions of nodes...
Any other solutions considered?

By default user should care about the index and make sure it is correct, this approach is just to make sure the conversion job will not crash or hang. Converter will show warning log for users to identify the correct index assigned.

My suggestion is that user just need to provide an array instead of map, and converter do `array->map` internally, or users can use TSV format as raw graph file.

mortid0: > Where are tests for feature fetching with the new ids?
> 
> The test cases are covered [here](https://github.com/microsoft/DeepGNN/blob/tiantiaw/update_converter/src/python/deepgnn/graph_engine/snark/tests/convert_test.py#:~:text=param%20%3D%20%5B,%5D).

I don't see any feature fetching in these tests.

> My suggestion is that user just need to provide an array instead of map, and converter do array->map internally, or users can use TSV format as raw graph file.

Why are we not doing it?

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -26,7 +35,12 @@ def triangle_graph_json(folder):
             "node_weight": 1,
             "neighbor": {"0": {"0": 0.5}, "1": {}},
             "uint64_feature": {},
-            "float_feature": {"0": [0, 1], "1": [-0.01, -0.02]},
+            "float_feature": [[0, 1], [-0.01, -0.02]]
```
mortid0, line 38: Please, [update documentation for proposed changes](https://github.com/microsoft/DeepGNN/blob/main/docs/advanced/data_spec.md#graph-data)

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
wangtt03: The purpose of this PR is try to simplify the user's converter job even when the raw graph is a wrong version, however, when I start doing this PR, I also noticed that maybe it is not a good idea to have a hidden id to replace the error id user provided. It is a tradeoff and the disadvantage of this PR is much bigger than the benefit. Actually, user should guarantee the correctness of the raw graph file, if they don't want to use json, they can use TSV because in TSV there is no need to generate such feature ids. Let me revert this PR, thanks.

PR #37, opened 2022-06-28
Move convert.py output fn to converter/process.py. Rename json_converter.py -> writers.py.

coledie:
* Make convert.py output fn private in its own file.
* Make the argument "process" in dispatchers optional.
* Move dispatcher flags to new process file.
* Rename json_converter.py -> writers.py.


src/python/deepgnn/graph_engine/snark/converter/process.py
```diff
@@ -0,0 +1,131 @@
...
+
+class _NoOpWriter:
+    def add(self, _: typing.Any):
+        return
+
+    def close(self):
+        return
+
+
+def converter_process(
```
mortid0, line 37: This function doesn't look private despite of the title of this PR.

src/python/deepgnn/graph_engine/snark/BUILD
```diff
@@ -16,7 +16,8 @@ py_library(
         "convert.py",
         "meta_merger.py",
         "converter/__init__.py",
-        "converter/json_converter.py",
+        "converter/process.py",
```
mortid0, line 19: this is a breaking change, could you add a note in the changelog.md?

coledie: The breaking change is the converter argument right? Users should never import the json_converter since it only has writers.

src/python/deepgnn/graph_engine/snark/converter/process.py
coledie: I moved it out of the same file as MultiWorkersConverters and into the folder where there is no other user facing code.

mortid0: Anyone can access this function via
```python
from deepgnn.graph_engine.snark.converter.process import converter_process
```
This is a public function, because there is [no leading underscore](https://peps.python.org/pep-0008/#method-names-and-instance-variables)

src/python/deepgnn/graph_engine/snark/BUILD
mortid0: no, you removed one public function.

> Users should never import the json_converter since it only has writers.

That's not how conventions work: https://peps.python.org/pep-0008/#method-names-and-instance-variables

src/python/deepgnn/graph_engine/snark/converter/process.py
coledie: I updated the PR description and changelog, the main goal is to remove it from the MultiWorkersConverter file since the function is only used by Dispatchers.

src/python/deepgnn/graph_engine/snark/dispatcher.py
```diff
@@ -70,8 +73,7 @@ def __init__(
                 bool,
             ],
             None,
-        ],
-        decoder_type: DecoderType,
+        ] = converter_process,
```
mortid0, line 76: Please, use None as default value for arguments.

src/python/deepgnn/graph_engine/snark/dispatcher.py
```diff
@@ -57,6 +59,7 @@ def __init__(
         folder: str,
         parallel: int,
         meta: str,
+        decoder_type: DecoderType,
```
mortid0, line 62: why argument reorder is needed?

mortid0 (approved):

src/python/deepgnn/graph_engine/snark/dispatcher.py
coledie: Reverted, only for QueueDispatcher where I do not want to give partition_fn a default value.

PR #39, opened 2022-06-29
Remove DecoderType. 

coledie:
* Remove DecoderType.
* Converter decoder_type -> decoder_class that takes Decoder class or obj.


src/python/deepgnn/graph_engine/snark/dispatcher.py
```diff
@@ -189,7 +188,7 @@ def __init__(
         num_partitions: int,
         meta: str,
         partion_func: typing.Callable[[str], int],
-        decoder_type: DecoderType,
+        decoder_class: typing.Any,
```
mortid0, line 191: why do you want to use any type, not an instance variable of subtype `Decoder`(which is in the docstring?

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -31,6 +19,9 @@ def decode(self, line: str):
         """Decode the line into a "node" object."""
 
 
+DecoderType = TypeVar("DecoderType", bound=Decoder)
+
```
mortid0, line 23: why do we need this? Can we just use `decoders.Decoder` for type hints?

src/python/deepgnn/graph_engine/snark/dispatcher.py
```diff
@@ -85,8 +86,8 @@ def __init__(
             folder (str): Location of graph files.
             parallel (int): Number of parallel process to use for conversion.
             meta (str): Meta data about graph.
-            process (typing.Callable[ [typing.Union[mp.Queue, Connection], mp.Queue, str, int, int, int, DecoderType], None ]): Function to call for processing lines in a file.
-            decoder_type (DecoderType): Decoder which is used to parse the raw graph data file, Supported: json/tsv.
+            process (typing.Callable[ [typing.Union[mp.Queue, Connection], mp.Queue, str, int, int, int, Decoder], None ]): Function to call for processing lines in a file.
```
mortid0, line 89: Is it possible to stay consistent between type hints(DecoderType) and docstring(Decoder)?

src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -29,7 +30,7 @@ def __init__(
         graph_path: str,
         meta_path: str,
         output_dir: str,
-        decoder_type: DecoderType = DecoderType.JSON,
+        decoder: Optional[DecoderType] = None,
```
mortid0, line 33: why a new default value is `None`? I thought `PipeDispatcher` doesn't initialize it and there will be an exception. It seems this parameter can't be optional.

src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -210,9 +212,9 @@ def convert(self):
     parser.add_argument(
         "-t",
         "--type",
-        type=DecoderType,
-        default=DecoderType.JSON,
-        help="Type of the graph data file. Supported: json, tsv",
+        type=str,
+        default="json",
+        help="Type of the graph data file. Supported: linear, json, tsv",
```
mortid0, line 217: I don't think linear decoder have been introduced yet

src/python/deepgnn/graph_engine/snark/converter/process.py
```diff
@@ -96,7 +83,7 @@ def converter_process(
         if line == FLAG_ALL_DONE:
             break
 
-        node = decoder.decode(line)
+        node = decoder.decode(line)  # type: ignore
```
mortid0, line 86: why do we need to start ignoring types here and it didn't happen before?

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: No TypeVar is what typing provides to handle inheritance. Decoder is valid but does not handle inheritance.

src/python/deepgnn/graph_engine/snark/convert.py
coledie: TypeVar does not work with default arguments. The dispatchers just pass this value to the convert_process/output function which is where the "if decoder_type is None: decoder_type = JsonDecoder()".

src/python/deepgnn/graph_engine/snark/converter/process.py
coledie: Since decoder has to be optional to accept default values, this has to be added.

src/python/deepgnn/graph_engine/snark/convert.py
coledie: TypeVar is the only way to handle inheritance with python typing. The optional is required to support the default value with a TypeVar.

coledie: Default value now only for MultiWorkerConverter not dispatchers.

mortid0 (approved):

PR #40, opened 2022-06-29
gitignore update setup.py outputs. Add tutorials outputs.

coledie:


.gitignore
```diff
@@ -8,10 +8,14 @@ bazel-DeepGNN*
 bazel-testlogs*
 
 # Python package artifacts
-dist/*
-build/*
+src/python/dist/*
+src/python/build/*
 *.egg-info
 
+# Examples output
```
mortid0, line 15: we shouldn't run examples code outside of bazel, because we don't want to be state dependent.

PR #42, opened 2022-06-30
Simplify decoder interface and optimize binary writers.

coledie:
* Make Decoder.decode a generator that yields a node then its outgoing edges in order. Yield format is (-1/src, node_id/dst, type, weight, features), with features being a list of dense features as ndarrays and sparse features as 2 tuples, coordinates and values.
* Reduce a lot of work to create "node" objects in decoders then unravel them in writers.
* Split up writers to support node and edges being yielded separately - but still require node first then its outgoing edges.
* Update return type for features in a way that is almost ready to write to file.
* Writers reduce file.tell() usage since is slow - replace with counters.

### DIFF
--- a/src/python/deepgnn/graph_engine/snark/converter/process.py
+++ b/src/python/deepgnn/graph_engine/snark/converter/process.py
@@ -21,7 +21,7 @@ PROCESS_PRINT_INTERVAL = 1000
 
 
 class _NoOpWriter:
-    def add(self, _: typing.Any):
+    def add(self, *_: typing.Any):
         return
 
     def close(self):
@@ -61,7 +61,8 @@ def converter_process(
     node_type_count = [0] * node_type_num
     edge_weight = [0] * edge_type_num
     edge_type_count = [0] * edge_type_num
-    writer = writers.NodeWriter(str(folder), suffix)
+    node_writer = writers.NodeWriter(str(folder), suffix)
+    edge_writer = writers.EdgeWriter(str(folder), suffix)
     node_alias: typing.Union[writers.NodeAliasWriter, _NoOpWriter] = (
         _NoOpWriter()
         if skip_node_sampler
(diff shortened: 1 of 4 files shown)


src/python/deepgnn/graph_engine/snark/converter/process.py
```diff
@@ -83,21 +84,23 @@ def converter_process(
...
+        for src, dst, typ, weight, features in decoder.decode(line):  # type: ignore
+            if src == -1:
+                node_writer.add(dst, typ, features)
+                edge_writer.add_node()
+                node_alias.add(dst, typ, weight)
+                node_weight[typ] += float(weight)
+                node_type_count[typ] += 1
+                node_count += 1
+            else:
+                edge_writer.add(dst, typ, weight, features)
```
mortid0, line 96: where do we verify edges are sorted by dst? without ordering the output of edge writer will be garbage. Currently it is done in NodeWriter, it records an offset to all edges belonging to it. Here you take lines from a queue and ordering is not guaranteed: a node data might be sent to worker processing partition 1, but some of its edges might go to a worker assigned a partition 2.

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -43,26 +41,30 @@ def __init__(self, folder: str, partition: int):
...
-            node (typing.Any): dictionary with information about
+            node_id: int
+            node_type: int
+            features: list[ndarray]
         """
-        self.nm.write(ctypes.c_uint64(node["node_id"]))  # type: ignore
-        self.nm.write(ctypes.c_uint64(self.count))  # type: ignore
-        self.nm.write(ctypes.c_int32(node["node_type"]))  # type: ignore
-        self.feature_writer.add(node)
-        self.edge_writer.add(node)
```
mortid0, line 58: This line is extremely important, node offset has to point to the specific location in the edge file. If you are separating nodes from edges, then there should be another way to keep them consistent, but I don't see it in the proposed changes.

mortid0 (changes requested):
It seems you are replacing a very simple interface for binary conversion of `NodeWriter` with an extremely complex function `converter_process`. It was created as a utility function, not as an entry point for conversion and this change creates lots of corner cases. Since we fill mp.Queues with edges rather than nodes, then edge ordering across workers producing binary files is not stable and edges with the same source can appear in every partition.

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -17,16 +16,97 @@
...
+        self.edge_type_count = [0] * self.edge_type_num
+
+    def add(self, data: typing.Iterator[typing.Tuple[int, int, int, float, list]]):
+        """
+        Write binary for a node and all of its edges.
+
+        args:
+            data: Iterable[(int, int, int, float, list)] Data for a node first, then all of
+                its edges in order of dst. Each entry for a node/edge is,
+                (-1/src, node_id/dst, type, weight, [ndarray for each feature vector or None in order of feature index]).
```
mortid0, line 82: how would you sort a following graph file with two nodes 2 and 3 connected to each other?
```csv
-1 2 0 1 [1,1,1]
-1 3 0 1 [2,2,2]
 2 3 0 1 []
 3 2 0 1 []
 ```
 It seems like it requires some manual reordering and for large graphs I can't think of any sorting procedure that will organize graph items in the required order.

Comment (coledie) [reply src/python/deepgnn/graph_engine/snark/converter/writers.py:82]: Right now,
```
  -1 2 0 1 [1,1,1]
  2 3 0 1 []
  -1 3 0 1 [2,2,2]
  3 2 0 1 []
```
In the future when nodes/edges do not need to be on the same partition it would not need to change.

coledie: I could write a function from this file to a buffer file that would be the one to get converted, if the original file is sorted all nodes then all edges the complexity would not be bad.

mortid0 (approved):

PR #43, opened 2022-06-30
Resolve all outstanding mypy issues - mostly np.array -> np.ndarray.

coledie:
These were preventing me from using pre-commit install.


mortid0:
Could you update `mypy.ini` to show which issues are resolved? It seems it is empty, do we need to update mypy?

coledie:
No update needed to mypy.ini.

mortid0:
> No update needed to mypy.ini.

Then how these issues can be triggered and why we don't see them in CI pipeline?

coledie:
I use "pre-commit run --all-files" a lot which is where I get these.

mortid0:
> I use "pre-commit run --all-files" a lot which is where I get these.

Probably CI is using [an old version of mypy](https://github.com/microsoft/DeepGNN/blob/main/.github/actions/lint/action.yml#L15) in comparison your local, could you update in this PR, so we'll catch these issues later?

PR #46, opened 2022-07-05
Remove FeatureType enum, replace with np.dtype.

coledie:

New Behavior
----------------
* Remove FeatureType enum, replace with np.dtype. 
* Fetch features parameter feature_type -> dtype.


src/python/deepgnn/pytorch/common/args.py
```diff
@@ -10,7 +10,7 @@
 def init_model_common_args(parser: argparse.ArgumentParser):
     """Model configuration."""
     group = parser.add_argument_group("Model Common Parameters")
-    group.add_argument("--feature_type", default="float", type=str, choices=["float", "uint64", "binary"], help="Feature type.")
+    group.add_argument("--dtype", default="float", type=str, choices=["float", "uint64", "binary"], help="Feature type.")
```
mortid0, line 13: please, keep argument name `feature_type`, it is more obvious than an obscure dtype reference.

src/python/deepgnn/pytorch/modeling/base_model.py
```diff
@@ -20,25 +20,25 @@ class BaseModel(nn.Module):
 
     def __init__(
         self,
-        feature_type: FeatureType,
+        dtype: np.dtype,
```
mortid0, line 23: Please, keep original argument names, they are more descriptive. It is weird to see  label_id and label_type, but feature_id and dtype combinations for supervised models.

src/python/deepgnn/pytorch/common/utils.py
```diff
@@ -13,19 +13,18 @@
 
 from deepgnn import get_logger
 from deepgnn.pytorch.common.consts import PREFIX_CHECKPOINT
-from deepgnn.graph_engine import FeatureType
 
 
-def get_feature_type(feature_type_str: str):
+def get_dtype(dtype_str: str):
```
mortid0, line 18: Can we get rid of this function completely and parse arguments with a plain `np.dtype` constructor?

coledie: It converts form np.dtype to python type - I cannot remove its use case. And the function is a bit complex: type(dtype(0).item()), maybe better to just have the helper function.

mortid0 (approved):

src/python/deepgnn/migrate.py
```diff
@@ -0,0 +1,44 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Migrate to new versions of DeepgNN.")
+    parser.add_argument(
+        "--script_dir", type=str, required=True, help="Directory to migrate."
+    )
+    args = parser.parse_args()
+
+    for filename in Path(args.script_dir).glob("**/*.py"):
+        if filename.name.endswith("migrate.py"):
```
mortid0, line 20: Why is this needed? We are only migrating external repos, right?

src/python/deepgnn/migrate.py
```diff
@@ -0,0 +1,44 @@
+"""
```
mortid0, line 1: Create a new folder for this file, called `migrate` and rename file `to_0_1_56.py`. This way users can run upgrades across multiple versions.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+We provide a python module to help you upgrade your scripts to new deepgnn versions.
```
mortid0, line 4: Put this to readme instead. A similar test is already present below.

src/python/deepgnn/migrate.py
coledie: I think we should just leave the migrate.py file, if user installs pip package 0.1.56, the migrate script will pull them up to that version, if a user wants a migrate script for a different version, they can just pip install that version.

src/python/deepgnn/migrate.py
coledie: I used it on DeepGNN since it is easier. I'll make it a full path so its moer safe.

src/python/deepgnn/migrate.py
mortid0: How can you handle migration between any possible combination of version ranges? Are you going to put everything in one giant file and introduce extra arguments for each version?

src/python/deepgnn/migrate.py
mortid0: But we do want to migrate users code if it is called migrate.py too, right?

src/python/deepgnn/migrate.py
coledie: We can add a start-version parameter so they can input the initial version of the code. I think that will make it easier than running multiple files. This change it does not matter though, you can rerun the migrate script and it wont break.

src/python/deepgnn/migrate.py
coledie: I'll change to src/python/deepgnn/migrate.py.

src/python/deepgnn/migrate.py
coledie: We can section the versions with comments. What is skip and override? If it fails half way through user will have to reset code and migrate again.

mortid0: Think about migration as stacked diffs/patches: split a big script into atomic changes and choose which ones you want to apply. Instead of a version control system, you have a python script to make transformations. If some changes are only relevant to e.g. python upgrades to async code, then an external user probably already applied and would rather skip them (simply delete relevant file). This structure is useful if there is a range of changes to apply and something must be skipped.

coledie: I see what you are saying theoretically, but there should never be a real scenario where the user wants to skip changes correct? There will be scenarios where changes just dont get applied cause user never uses FeatureType or something. If we have one user looking for advanced usage they can copy the script and delete the relevant sections quickly. Maybe we can split the migrate script into multiple files when/if it is necessary?

mortid0: > there should never be a real scenario where the user wants to skip changes correct?

Hard to say, it is definitely not 0, because I am always surprised how code is used in the wild. I'd rather err on the side of caution, especially since all is needed from us is to move a file to a separate folder.

src/python/deepgnn/migrate/0_1_56.py
```diff
@@ -0,0 +1,44 @@
+"""
```
mortid0, line 1: where is copyright?

src/python/deepgnn/migrate/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""DeepGNN modules to work with torch models."""
```
mortid0, line 3: That's definitely a wrong docstring.

src/python/deepgnn/migrate/0_1_56.py
```diff
@@ -0,0 +1,44 @@
...
+Script to migrate to new versions of DeepGNN.
+
+pip install google-pasta
```
mortid0, line 4: why is this needed in the docstring?

src/python/deepgnn/migrate/0_1_56.py
```diff
@@ -0,0 +1,44 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Migrate to new versions of DeepgNN.")
+    parser.add_argument(
+        "--script_dir", type=str, required=True, help="Directory to migrate."
+    )
+    args = parser.parse_args()
+
+    for filename in Path(args.script_dir).glob("**/*.py"):
+        if filename.name.startswith("src/python/deepgnn/migrate"):
```
mortid0, line 20: Again, this is not needed.

CHANGELOG.md
```diff
@@ -11,9 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Breaking. Added version checks for binary data. Requires to convert graph data or add v1 at the top of meta files.
 
+- Add migrate script to pull to new version of deepgnn.
+
 - Add debug mode to MultiWorkersConverter, using debug=True will now disable multiprocessing and show error messages.
 
 ### Changed
+- Remove FeatureType enum, replace with np.dtype. FeatureType.BINARY -> np.uint8, FeatureType.FLOAT -> np.float32, FeatureType.INT64 -> np.int64.
```
mortid0, line 19: Add `Breaking` in front.

README.md
```diff
@@ -27,3 +27,13 @@ cd examples/pytorch/graphsage
 ### Training other models
 
 Examples folder contains various models one can experiment with DeepGNN. To train models with Tensorflow you need to install `deepgnn-tf` package, `deepgnn-torch` package contains packages to train pytorch examples. Each model folder contains a shell script `run.sh` that will train a corresponding model on a toy graph, a `README.md` file with a short description of a model, reference to original paper, and explanation of command line arguments.
+
+## Migrating Scripts
+
+We provide a python module to help you upgrade your scripts to new deepgnn versions.
+
+```bash
+python -m deepgnn.migrate.0_1_56 --script_dir .
```
mortid0, line 36: replace dot with a comment `directory to migrate`.

PR #44, opened 2022-07-01
Neighbor Count Functionality for Memory/Distributed Graphs

adithyabhonsley:
- [X ] Forked repo is synced with upstream -> github shows no code delta outside of the desired.
    https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork
- [X ] Tests are passing? https://github.com/microsoft/DeepGNN/blob/main/CONTRIBUTING.md#run-tests
- [X ] Documentation is added or updated to reflect new code in the same format as the rest of the repo.
- [ X] PR is labeled using the label menu on the right side.

Previous Behavior
----------------
* Provide relevant issue number if applicable, eg #44.

New Behavior
----------------
* Adds neighbor count (node degree) functionality for both memory graph and distributed graphs
* Adds testing for neighbor count feature for memory/distributed graph (base cases, single/multiple partition, invalid nodes and types)
* Adds Python handling of c functions leveraging ctypes library

PR #50, opened 2022-07-14
[BUG FIX]: Dropped Indexing for 2-hop neighbor_feats

Swapnil-Gandhi:
Bug Description: Previously, indices `idx` (output of numpy.unique) of 1-hop neighbors (i.e. `neigh_feats_unique` computed on Line 83) were being used to reconstruct features of 2-hop neighbors (i.e. `neigh_feats_unique["neighbor_feats"]` computed on Line 90); this results in slicing of feature matrix for 2-hop neighbors.

This change drops the indexing on Line 90.

Steps to reproduce:
(Dataset: Cora; Fanout: [10, 10]; Batch-Size: 140; Sampling: RandomWithoutReplacement)

Before this code change:

```
>>> context['encoder']['node_feats']['node_feats'].shape
(140, 50)

>>> context['encoder']['node_feats']['neighbor_feats'].shape
(1400, 50)

>>> context['encoder']['neighbor_feats']['node_feats'].shape
(1400, 50)

>>> context['encoder']['neighbor_feats']['neighbor_feats'].shape
(1400, 50)
```

After this code change:

```
>>> context['encoder']['node_feats']['node_feats'].shape
(140, 50)

>>> context['encoder']['node_feats']['neighbor_feats'].shape
(1400, 50)

>>> context['encoder']['neighbor_feats']['node_feats'].shape
(1400, 50)

>>> context['encoder']['neighbor_feats']['neighbor_feats'].shape
(4160, 50)
```

- [ ] Forked repo is synced with upstream -> github shows no code delta outside of the desired.
- [ ] Documentation is added or updated to reflect new code in the same format as the rest of the repo.

Previous Behavior
----------------
* Provide relevant issue number if applicable, eg #44.

New Behavior
----------------

### DIFF
--- a/src/python/deepgnn/pytorch/encoding/gnn_encoder_sage.py
+++ b/src/python/deepgnn/pytorch/encoding/gnn_encoder_sage.py
@@ -87,7 +87,7 @@ class SageEncoder(nn.Module):
         if isinstance(neigh_feats_unique, dict):
             context["neighbor_feats"] = {
                 "node_feats": neigh_feats_unique["node_feats"][idx],
-                "neighbor_feats": neigh_feats_unique["neighbor_feats"][idx],
+                "neighbor_feats": neigh_feats_unique["neighbor_feats"],
                 "node_count": idx.size,
             }
         else:


coledie:
Hi Swapnil, good catch on this bug! Though I think the solution is not proper, instead neighbors shape should be different.

Before line 90(line you edited), the shapes are,
(414, 50) (4140, 50)

But they should be,
(414, 50) (414, 10, 50)

So that after line 90 they are,
(1400, 50) (1400, 10, 50)
This way everything is still properly upsampled without the information loss you found.

Though a simple reshape does not fix this, since it errors in the aggregator. I will continue to work on a solution for this tomorrow, I think there are two bugs that cancel each other out to make the code pass.

coledie:
Will open new PR and reference.

PR #57, opened 2022-07-21
Dispatcher bucket lines and send buckets through queue.

coledie:

New Behavior
----------------
Dispatcher now buckets lines and sends buckets through queue instead of sending each line individually. This affects some unit tests because now nodes data[:len(data) // n_partitions] are in the same partition instead of data[::n_partitions].

10M Edges, Single partition, JSON format:
original - 170s
externalize - 103s
This PR - 92s


src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -113,9 +113,11 @@ def convert(self):
             num_workers=self.worker_count,
         )
 
-        for _, data in enumerate(dataset):
-            for line in data:
-                d.dispatch(line)
+        for data in dataset:
+            data_len = len(data)
+            split_len = data_len // self.thread_count + 1
```
mortid0, line 118: how would you handle a case when `data_len` = 1 and `thread_count` = 1? `split_len` will be 0, right?

coledie: It will be 1 with the +1 at the end.

coledie:
Close in favor of removing dispatcher entirely.

PR #58, opened 2022-07-25
Support nodes and their outgoing edges on different partitions.

coledie:

New Behavior
----------------
* Support nodes and their outgoing edges on different partitions.
* Update node feature fetching to be more like edge feature fetching.
* Add test case to e2e test.


CHANGELOG.md
```diff
@@ -11,3 +11,5 @@
 * Add BinaryWriter as new entry point for NodeWriter, EdgeWriter and alias writers.
 
 * Meta.json files are no longer needed by the converter. Remove meta path argument from MultiWorkerConverter and Dispatchers.
+
+* Support nodes and their outgoing edges on different partitions.
```
mortid0, line 15: What does this mean? I don't see any changes to edges, only node features were updated.

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -77,9 +77,11 @@ def add(self, data: typing.Iterator[typing.Tuple[int, int, int, float, list]]):
                 its edges in order of dst. Each entry for a node/edge is,
                 (node_id/src, -1/dst, type, weight, [ndarray for each feature vector or None in order of feature index]).
         """
+        prev_node_id = -1
```
mortid0, line 80: Please, reduce number of use magic constants and use `None` or some other way to track previous node.

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -261,10 +267,11 @@ def add_node(self):
             ctypes.c_uint64(self.ei.tell() // (4 + 8 + 8 + 4))
         )  # 4 bytes type, 8 bytes destination, 8 bytes offset, 4 bytes weight
 
-    def add(self, dst: int, tp: int, weight: float, features: list):
+    def add(self, src: int, dst: int, tp: int, weight: float, features: list):
```
mortid0, line 270: `src` argument is not used, please remove it.

src/cc/lib/graph/graph.cc
```diff
@@ -125,9 +125,14 @@ void Graph::GetNodeFeature(std::span<const NodeId> node_ids, std::span<snark::Fe
         }
```
mortid0, line 125: Please, add a test with a request to node features for:
1. Node is not present in the graph. Expected dense feature vector should be 0.
2. Node is present in the graph, but feature is not present in any of partitions. FV should be 0s as well.

src/cc/lib/graph/partition.cc
```diff
@@ -176,7 +176,7 @@ void Partition::ReadEdgeIndex(std::filesystem::path path, std::string suffix)
 
     // Extra padding to simplify edge type count calculations.
     m_neighbors_index.back() = m_edge_types.size();
-    m_neighbors_index.push_back(m_edge_types.size());  // Fix for 0 edges in partition
+    m_neighbors_index.push_back(m_edge_types.size()); // Fix for 0 edges in partition
```
mortid0, line 179: Please, target main branch.

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -93,13 +95,17 @@ def add(self, data: typing.Iterator[typing.Tuple[int, int, int, float, list]]):
                 self.node_type_count[typ] += 1
                 self.node_count += 1
             else:
+                if src != prev_node_id:
+                    self.node_writer.add(src, -1, [])
```
mortid0, line 99: how do we identify a correct node type between multiple partitions then?

src/cc/lib/graph/partition.h
```diff
@@ -26,13 +26,13 @@ struct Partition
     Partition(std::filesystem::path path, std::string suffix, PartitionStorageType storage_type);
 
     Type GetNodeType(uint64_t internal_node_id) const;
-    void GetNodeFeature(uint64_t internal_node_id, std::span<snark::FeatureMeta> features,
+    bool GetNodeFeature(uint64_t internal_node_id, std::span<snark::FeatureMeta> features,
```
mortid0, line 29: Please, add C++ unit tests.

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
```diff
@@ -231,39 +247,120 @@ def __call__(self, x):
...
+            value["weight"],
+            get_features(value),
+        )
+
+
+def write_multi_binary(output_dir, partitions):
+    partition_meta = ""
+    for i, p in enumerate(partitions):
+        writer = BinaryWriter(output_dir, i)
+        writer.add([linearize(v) for v in p])
```
mortid0, line 298: That's not how `BinaryWriter` is supposed to work. From documentation:
```
data: Iterable[(int, int, int, float, list)] Data for a node first, then all of
                its edges in order of dst. Each entry for a node/edge is
```

There is nothing about splitting features and expected outcomes.

mortid0 (changes requested):

CHANGELOG.md
coledie: When an edge is added to a partition without its source node added first, this PR will add a dummy node with type = -1 - this way we support having a nodes outgoing edges on different partitions.

src/python/deepgnn/graph_engine/snark/converter/writers.py
coledie: We do not want any of these added nodes to have valid types, just ghost nodes so they get added to the node map and neighbor index.

mortid0: I don’t see any changes to GetNodeType function. This pr will return ghost types along with the regular node types.

mortid0: https://github.com/microsoft/DeepGNN/blob/7c573e532642a680b7be0361443c8868366e7bb2/src/cc/lib/distributed/graph_engine.cc#L86

CHANGELOG.md
mortid0: That's not how does [the decoder interface work right now](https://github.com/microsoft/DeepGNN/blob/main/src/python/deepgnn/graph_engine/snark/decoders.py#L36) :

> this is a generator that yields a node then its outgoing edges in order.

This means all edges will end up in the same partition. It seems like you are missing changes in the converter interface for this PR.

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
coledie: Updated this usage.

CHANGELOG.md
coledie: This is a documentation issue, I will fix.

src/python/deepgnn/graph_engine/snark/converter/writers.py
coledie: Added support for remote and local

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -34,6 +34,7 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
         """Decode the line of text.
 
         This is a generator that yields a node then its outgoing edges in order.
+        A node and its outgoing edges can be on different partitions.
```
mortid0, line 37: This is a non-trivial feature. Could you add an update documentation(the one in the doc folder) on how/when to use it?

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -93,6 +95,10 @@ def add(self, data: typing.Iterator[typing.Tuple[int, int, int, float, list]]):
                 self.node_type_count[typ] += 1
                 self.node_count += 1
             else:
+                if src != self.prev_node_id or self.prev_node_id is None:
+                    self.node_writer.add(src, -1, [])
```
mortid0, line 99: Please, update documentation, that -1 is the reserved node type and can't be used in a graph.

src/cc/lib/graph/partition.cc
coledie: In e2e test I added datasets where one partition has only edges and other partition has only nodes. It fails when .neighbors is run on this dataset.

src/cc/lib/distributed/graph_engine.cc
```diff
@@ -81,9 +81,15 @@ grpc::Status GraphEngineServiceImpl::GetNodeTypes(::grpc::ServerContext *context
             continue;
         }
 
-        const auto index = elem->second;
+        auto index = elem->second;
+        size_t partition_count = m_counts[index];
+        int32_t result = -1;
+        for (size_t partition = 0; partition < partition_count && result == -1; ++partition, ++index)
```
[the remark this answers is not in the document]
coledie: This is for tracking whether or not the type is found in the partition so it does not get overwritten. But I will make it so offset does not get added if type is not found.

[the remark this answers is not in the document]
mortid0: You add dummy types to response on line 92, even when `result=-1`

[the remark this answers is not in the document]
coledie: Fixed.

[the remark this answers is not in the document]
coledie: Removed

src/cc/lib/distributed/graph_engine.cc
```diff
@@ -110,13 +118,25 @@ grpc::Status GraphEngineServiceImpl::GetNodeFeatures(::grpc::ServerContext *cont
             continue;
         }
 
-        response->add_offsets(node_offset);
         response->mutable_feature_values()->resize(feature_offset + fv_size);
```
mortid0, line 121: Avoid multiple resizes, here and on line 138. Append features if and only if they are present in the partition.

src/cc/lib/distributed/graph_engine.cc
```diff
@@ -110,13 +118,25 @@ grpc::Status GraphEngineServiceImpl::GetNodeFeatures(::grpc::ServerContext *cont
...
+        size_t partition_count = m_counts[index];
         auto data = reinterpret_cast<uint8_t *>(response->mutable_feature_values()->data());
-        m_partitions[m_partitions_indices[index]].GetNodeFeature(m_internal_indices[index], features,
-                                                                 std::span(data + feature_offset, fv_size));
-        feature_offset += fv_size;
+        bool found = false;
+        for (size_t partition = 0; partition < partition_count && !found; ++partition, ++index)
+        {
+            found = m_partitions[m_partitions_indices[index]].GetNodeFeature(m_internal_indices[index], features,
+                                                                             std::span(data + feature_offset, fv_size));
```
mortid0, line 129: nit: take this variable `std::span(data + feature_offset, fv_size)` outside of the for loop.

src/cc/lib/distributed/graph_engine.cc
```diff
@@ -275,10 +300,15 @@ grpc::Status GraphEngineServiceImpl::GetNodeStringFeatures(::grpc::ServerContext
...
-        m_partitions[m_partitions_indices[index]].GetNodeStringFeature(
-            m_internal_indices[index], features, dimensions.subspan(features_size * node_offset, features_size),
-            values);
+        auto index = internal_id->second;
+        size_t partition_count = m_counts[index];
+        bool found = false;
+        for (size_t partition = 0; partition < partition_count && !found; ++partition, ++index)
+        {
+            found = m_partitions[m_partitions_indices[index]].GetNodeStringFeature(
+                m_internal_indices[index], features, dimensions.subspan(features_size * node_offset, features_size),
```
mortid0, line 309: nit: same as above, `dimensions.subspan(features_size * node_offset, features_size)` doesn't change in the loop and can be created outside.

mortid0: same below

src/cc/tests/distributed_test.cc
```diff
@@ -610,6 +610,214 @@ TEST(DistributedTest, FullNeighborsMultipleServers)
...
+    std::vector<std::shared_ptr<grpc::Channel>> channels;
+
+    TestGraph::MemoryGraph m1;
+    m1.m_nodes.push_back(TestGraph::Node{.m_id = 0, .m_type = 0, .m_weight = 1.0f, .m_float_features = f0});
+    m1.m_nodes.push_back(TestGraph::Node{.m_id = 1, .m_type = 1, .m_weight = 1.0f, .m_float_features = f1});
+    m1.m_nodes.push_back(TestGraph::Node{.m_id = 2, .m_type = -1, .m_weight = 1.0f});
+    TestGraph::MemoryGraph m2;
+    m2.m_nodes.push_back(TestGraph::Node{.m_id = 1, .m_type = -1});
+    m2.m_nodes.push_back(TestGraph::Node{.m_id = 2, .m_type = 2, .m_float_features = f2});
+    std::vector<TestGraph::MemoryGraph> test_graphs = {m1, m2};
```
mortid0, line 705: Please, add one more case with `m1` and `m2` swapped to verify we correctly skip partitions before a feature is found.

src/cc/tests/distributed_test.cc
```diff
@@ -610,6 +610,214 @@ TEST(DistributedTest, FullNeighborsMultipleServers)
...
+}
+
+TEST(DistributedTest, NodeStringFeaturesMultipleTypesNeighborsSpreadAcrossPartitions)
+{
+    std::vector<std::vector<float>> f0 = {std::vector<float>{1.0f, 2.0f, 3.0f}};
+    std::vector<std::vector<float>> f1 = {std::vector<float>{4.0f, 5.0f, 6.0f}};
+    std::vector<std::vector<float>> f2 = {std::vector<float>{7.0f, 8.0f, 9.0f}};
+
+    auto environment = CreateMultiServerSplitFeaturesEnvironment(
+        "NodeFeaturesMultipleTypesNeighborsSpreadAcrossPartitions", f0, f1, f2);
```
mortid0, line 768: nit: use test name as environment name to store binary data in separate folders, `NodeStringFeaturesMultipleTypesNeighborsSpreadAcrossPartitions`, same below.

mortid0: same below

src/cc/tests/distributed_test.cc
coledie: Yes this is already the case, in m1, id = 1 is a real node and id = 2 is a ghost node.

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: I dont think there is any case where someone would go out of their way to use this, but it is important for people to know so they dont avoid it. So I think it is best that we document it on the decoder and BinaryWriter docstings?

src/cc/lib/distributed/graph_engine.cc
coledie: We need to resize it beforehand so if the feature does exist, it'll fit. We only do the second resize if the feature is not found. This is a copy of the GetEdgeFeatures code which also has 2.

src/cc/lib/graph/partition.cc
mortid0: This means it is a bug in the converter, not in C++ code: node map can't contain an internal index pointing to the last element in the neighbor index, because we add an extra element in the close method.

src/cc/lib/distributed/graph_engine.cc
```diff
@@ -81,9 +81,17 @@ grpc::Status GraphEngineServiceImpl::GetNodeTypes(::grpc::ServerContext *context
             continue;
         }
 
-        const auto index = elem->second;
+        auto index = elem->second;
+        size_t partition_count = m_counts[index];
```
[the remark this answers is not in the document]
mortid0: sorry, `const size_t partition_count = m_counts[index];`

src/cc/lib/graph/partition.cc
coledie: Removed, looks like I resolved this when updating theBinaryWriter.add

src/cc/lib/graph/graph.cc
```diff
@@ -99,8 +99,14 @@ void Graph::GetNodeType(std::span<const NodeId> node_ids, std::span<Type> output
...
         else
         {
-            const auto index = internal_id->second;
-            *curr_type = m_partitions[m_partitions_indices[index]].GetNodeType(m_internal_indices[index]);
+            auto index = internal_id->second;
+            size_t partition_count = m_counts[index];
+            for (size_t partition = 0; partition < partition_count; ++partition, ++index)
+            {
+                *curr_type = m_partitions[m_partitions_indices[index]].GetNodeType(m_internal_indices[index]);
```
mortid0, line 106: Extract a small function `HasNodeFeatures(intenrnal_node_id) -> bool` and use it instead of copying the same cryptic call to `GetNodeType`.

src/cc/tests/distributed_test.cc
```diff
@@ -610,6 +610,234 @@ TEST(DistributedTest, FullNeighborsMultipleServers)
...
+            std::make_shared<snark::GraphEngineServiceImpl>(path.string(), std::vector<uint32_t>{0},
+                                                            snark::PartitionStorageType::memory, ""),
+            std::shared_ptr<snark::GraphSamplerServiceImpl>{}, "localhost:0", "", "", ""));
+        channels.emplace_back(servers.back()->InProcessChannel());
+    }
+
+    return std::make_pair(std::move(servers), std::make_shared<snark::GRPCClient>(std::move(channels), 1, 1));
+}
+
+TEST(DistributedTest, NeighborSampleMultipleTypesNeighborsSpreadAcrossPartitions)
```
mortid0, line 653: Why do we need new neighbor sampling tests? I thought this PR is only for features.

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
```diff
@@ -647,10 +821,11 @@ def test_remote_client_node_features_single_server(
     "storage_type",
     [client.PartitionStorageType.memory, client.PartitionStorageType.disk],
 )
+@pytest.mark.parametrize("multi_partition_graph_data", param, indirect=True)
 def test_distributed_graph_metadata(multi_partition_graph_data, storage_type):
     address = ["localhost:2234", "localhost:2235"]
     if platform.system() != "Darwin":
-        address = [a + f"{int(storage_type)}" for a in address]
+        address = [a + f"{int(storage_type)}{int(time.time()) % 100}" for a in address]
```
mortid0, line 828: why this change is needed?

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
```diff
@@ -24,87 +24,103 @@
...
+        "int8_feature": {"5": [15, 16, 17]},
+        "uint16_feature": {"6": [25, 26, 27]},
+        "int16_feature": {"7": [35, 36, 37]},
+        "uint32_feature": {"8": [45, 46, 47]},
+        "int32_feature": {"9": [55, 56, 57]},
+        "uint64_feature": {"10": [65, 66, 67]},
+        "int64_feature": {"11": [75, 76, 77]},
+        "double_feature": {"12": [85, 86, 87]},
+        "float16_feature": {"13": [95, 96, 97]},
+        "sparse_float16_feature": {
```
mortid0, line 107: We test sparse features in a separate suite. Could you update `sparse_features_test.py`?

mortid0:
I think the readthedocs failure means you need to rebase the PR.

mortid0 (approved):

PR #59, opened 2022-07-27
Add Linear format.

coledie:

New Behavior
----------------
* Add a new converter input format called "Linear" with nodes and edges on separate lines.
* Built in datasets now use linear for speed gains.
* Add to unit tests - convert_tests, e2e_tests and sparse_features_test.

Notes
-------
* No checks for if the file is sorted properly or not.
* Linear is similar to the TSV format, but different enough where they cant really be merged(neighbor edges are on the same line as node, feature encoding).


src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -488,7 +532,9 @@ def graph_with_sparse_features_json(folder):
             "node_type": 1,
             "node_weight": 1,
             "float_feature": {"0": [1], "1": [-0.03, -0.04]},
-            "sparse_float_feature": {"2": {"coordinates": [1, 3, 7], "values": [5.5]}},
+            "sparse_float_feature": {
+                "2": {"coordinates": [[1, 3, 7]], "values": [5.5]}
```
coledie, line 536: I updated this test case from,
```python
  "sparse_float_feature": {"2": {"coordinates": [1, 3, 7], "values": [5.5]}},
```
to
```python
  "sparse_float_feature": {"2": {"coordinates": [[1, 3, 7]], "values": [5.5]}}
```
it seems like the original version was incorrect?

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,186 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+        <edge_1_info>
+        <edge_2_info>
+        ...
+        ```
+        node_info: node_id -1 node_type node_weight <features>
+        edge_info: src dst edge_type edge_weight <features>
+        features[dense]: dtype_name length v1 v2 ... dtype_name2 length2 v1 v2 ...
+        features[sparse]: dtype_name values.size,coords.shape[1] c1 c2 ... v1 v2 ...
+        features[sparse]: dtype_name values.size,0 c1 c2 ... v1 v2 ...
+        * Nodes must be sorted by node_id, edges sorted by src and then dst.
```
mortid0, line 61: do edges have to be sorted by edge type too?

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,186 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+        the following keyword arguments can be added when creating the decoder.
+        default_node_type: int Type of all nodes, if set do not add node type to any nodes.
+        default_node_weight: int Weight of all nodes, if set do not add node weight to any nodes.
+        default_node_feature_types: ["dtype" or None, ...] Dtype of each feature vector.
+        default_node_feature_lens: [[int, ...] or None, ...] Length value for each feature vector.
+        default_edge_type: int Same as node except for all edges.
+        default_edge_weight: int Same as node except for all edges.
+        default_edge_feature_types: ["dtype" or None, ...] Dtype of each feature vector.
+        default_edge_feature_lens: [[int, ...] or None, ...] Length value for each feature vector.
+        ```
```
mortid0, line 85: what does this section(lines 86-103) represent?

src/cc/lib/graph/partition.cc
```diff
@@ -293,7 +293,7 @@ void Partition::ReadEdgeFeaturesData(std::filesystem::path path, std::string suf
 {
     if (is_hdfs_path(path))
     {
-        auto full_path = path / ("edge_features" + suffix + ".data");
+        auto full_path = path / ("edge_features_" + suffix + ".data");
```
mortid0, line 296: how is it related to linear format?

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
mortid0: we support both cases, a single list(in case of a single value) and a nested list. Your case is more explicit.

docs/graph_engine/data_spec.rst
```diff
@@ -17,9 +17,77 @@ Users can generate a graph in either format then our pipeline will convert it in
...
+	<node info>
+	<edge_1_info>
+	<edge_2_info>
+	...
+
+node_info: node_id -1 node_type node_weight <features>
+edge_info: src dst edge_type edge_weight <features>
+features[dense]: dtype_name length v1 v2 ... dtype_name2 length2 v1 v2 ...
+features[sparse]: dtype_name values.size,coords.shape[1] c1 c2 ... v1 v2 ...
+features[sparse]: dtype_name values.size,0 c1 c2 ... v1 v2 ...
```
mortid0, line 44: why there are 2 representations of sparse features? Could you add explanation to the doc?

docs/graph_engine/data_spec.rst
```diff
@@ -17,9 +17,77 @@ Users can generate a graph in either format then our pipeline will convert it in
...
+.. code-block:: text
+	<node info>
+	<edge_1_info>
+	<edge_2_info>
+	...
+
+node_info: node_id -1 node_type node_weight <features>
+edge_info: src dst edge_type edge_weight <features>
+features[dense]: dtype_name length v1 v2 ... dtype_name2 length2 v1 v2 ...
+features[sparse]: dtype_name values.size,coords.shape[1] c1 c2 ... v1 v2 ...
```
mortid0, line 43: what does coords.shape[1] mean?

mortid0 (commented):
Where are tests for a node is split across multiple partitions?

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
coledie: Okay I reverted this.

src/cc/lib/graph/partition.cc
coledie: Removed.

CHANGELOG.md
```diff
@@ -13,3 +13,5 @@
 * Meta.json files are no longer needed by the converter. Remove meta path argument from MultiWorkerConverter and Dispatchers.
 
 * Support nodes and their outgoing edges on different partitions.
+
+* Add new converter input format "Linear" with LinearDecoder. Format has nodes and edges on separate lines, is smaller and faster to convert.
```
mortid0, line 17: Unreleased features are added in the top: https://keepachangelog.com/en/1.0.0/

docs/graph_engine/data_spec.rst
```diff
@@ -17,9 +17,77 @@ Users can generate a graph in either format then our pipeline will convert it in
...
-1. `JSON <#json-format>`_: Heterogeneous or homegeneous graph.
+1. `Linear <#linear-format>`_: Heterogeneous or homegeneous graph.
 
-2. `TSV <#tsv-format>`_: Homogeneous graph only.
+2. `JSON <#json-format>`_: Heterogeneous or homegeneous graph.
+
+3. `TSV <#tsv-format>`_: Homogeneous graph only.
+
+Linear Format
+===========
```
mortid0, line 27: nit: check formatting, length should be equal to the string above.

docs/graph_engine/data_spec.rst
```diff
@@ -17,9 +17,77 @@ Users can generate a graph in either format then our pipeline will convert it in
...
+
+Linear format is most directly what is given to binary writers, it supports nodes and their outgoing edges in
```
mortid0, line 29: Please, avoid using new terminology/implementation detail right away from the start. What is binary writer and why it is important to describe a linear format?

docs/graph_engine/data_spec.rst
```diff
@@ -17,9 +17,77 @@ Users can generate a graph in either format then our pipeline will convert it in
...
+	<edge_1_info>
+	<edge_2_info>
+	...
+
+node_info: node_id -1 node_type node_weight <features>
+edge_info: src dst edge_type edge_weight <features>
+features[dense]: dtype_name length v1 v2 ... dtype_name2 length2 v1 v2 ...
+features[sparse with 2 dim coordinates vector]: dtype_name values.size,coords.shape[1] c1 c2 ... v1 v2 ...
+features[sparse with 1 dim coordinates vector]: dtype_name values.size,0 c1 c2 ... v1 v2 ...
+* Nodes must be sorted by node_id, edges sorted by src and then dst.
```
mortid0, line 45: Edges have to be sorted by type first and then by dst.

docs/graph_engine/data_spec.rst
```diff
@@ -17,9 +17,77 @@ Users can generate a graph in either format then our pipeline will convert it in
...
+
+.. code-block:: text
+	<node info>
+	<edge_1_info>
+	<edge_2_info>
+	...
+
+node_info: node_id -1 node_type node_weight <features>
+edge_info: src dst edge_type edge_weight <features>
+features[dense]: dtype_name length v1 v2 ... dtype_name2 length2 v1 v2 ...
```
mortid0, line 42: Could you add an explanation how feature ids will be assigned to nodes/edges? How to handle missing features, i.e. if some nodes should have only features with id 0 and 2, while others should have features with ids 0, 1 and 2

src/python/deepgnn/graph_engine/snark/tests/util_test.py
```diff
@@ -0,0 +1,77 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Unit test utils."""
```
mortid0, line 4: Can we use these mocks as fixtures instead and let pytest to cache them?

coledie: These are just functions, so not directly. Since they are used only within fixtures though they are cached.

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,191 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+    def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
+        """Convert text line into a node and edge iterator.
+
+        This is a generator that yields a node then its outgoing edges in order.
+        A node and its outgoing edges can be on different partitions.
+        Yield format is (node_id/src, -1/dst, type, weight, features).
+        Features being a list of dense features as ndarrays and sparse features as 2 tuples, coordinates and values.
+        """
+        if line == "":
+            return []
```
mortid0, line 153: why do we return, not yield an empty result?

coledie: This function is used in a for loop, so here on the null case we return an empty iterator and the for loop will not even run a single loop. If we yield a None, the for loop will at least run one loop on that and if it does not break it will attempt to loop again where it'll get an error.

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,191 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+
+        data = iter(line.split(","))
```
mortid0, line 155: line.split returns a list of strings, which uses lots of memory. Why can't we use a CSV reader?

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,191 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+                self.edge_feature_types,
+                self.edge_feature_lens,
+                self.n_edge_feature,
+            )
+        if typ is None:
+            typ = int(next(data))
+        if weight is None:
+            weight = float(next(data))
+
+        features = []
```
mortid0, line 177: could you extract feature parsing in a separate function? It is hard to follow through.

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,191 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+        n_features = 0
```
mortid0, line 178: why do we need to store list length separately?

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: I use this direclty in a for loop to stream the input, eg
for item in decoder.decode(line):
if I change this to
for item in list(decoder.decode(line)):
it will no longer stream

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: csv reader also goes line by line: 
https://github.com/python/cpython/blob/da4aae29f829fdf288b42919671274d7d463fe04/Lib/csv.py#:~:text=self.fieldnames-,row%20%3D%20next(self.reader),-self.line_num

Each line is only a single node or edge so the limit is very large feature vectors

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: Calling len() is slow and each loop calls it 3 times.

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Support nodes and their outgoing edges on different partitions.
 
+* Add new converter input format "EdgeList" with EdgeListDecoder. Format has nodes and edges on separate lines, is smaller and faster to convert.
+
+* Bugfix: fill dimensions with 0 for missing features.
```
mortid0, line 13: Please, fix format. See line 17 for example.

src/python/deepgnn/graph_engine/data/data_util.py
```diff
@@ -26,49 +25,30 @@ def download_file(url: str, data_dir: str, name: str):
...
-                "weight": 1.0,
-            }
-            for nb in test_neighbors
-        ],
-    }
-    return json.dumps(node)
+
+    output = ""
+    output += f"{node_id},-1,{ntype},1.0,float32,{len(flt_feat)},{','.join([str(v) for v in flt_feat])},float32,1,{label}\n"
+    for nb in train_neighbors:
```
mortid0, line 47: I don't think iteration over set produces a sorted output.

mortid0: Same for test_neighbors and ppi dataset.

src/python/deepgnn/graph_engine/snark/decoders.py
mortid0: I am talking about parsing lines, not reading them from files. 
The difference between str.split and CSV reader is the latter reader doesn't create new strings, but [updates C fields](https://github.com/python/cpython/blob/f62ad4f2c4214fdc05cc45c27a5c068553c7942c/Modules/_csv.c#L657) which is way faster.

src/python/deepgnn/graph_engine/snark/decoders.py
mortid0: Since there is no other reason, plese use `len` method. The bottleneck is using `str.split` method, not calling a function to return a number.

src/python/deepgnn/graph_engine/data/citation.py
```diff
@@ -264,48 +263,34 @@ def _write_json_graph(
...
-                    "weight": float(1),
-                    "edge_type": 0,
-                }
-                for nb in neighbors
-            ],
-        }
-        return json.dumps(node)
+        output = ""
+        output += f"{node_id},-1,{ntype},1.0,float32,{len(flt_feat)},{','.join([str(v) for v in flt_feat])},float32,1,{label}\n"
+        for nb in neighbors:
```
mortid0, line 291: 1. Use `sorted`.
2. Add a unit test that verifies neighbors are processed in the correct order.

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: Each line has a different length and we can't know the types beforehand which csv reader does not support.

src/python/deepgnn/graph_engine/data/citation.py
coledie: I think we should add something to the converter / loader to fail if they are not sorted properly. It is not easy to add a test in test_dataset.

src/python/deepgnn/graph_engine/snark/decoders.py
mortid0: Of course csv reader supports variable length and different types. Using example from the docs:
image

src/python/deepgnn/graph_engine/snark/decoders.py
mortid0: I just didn't want to print a for loop in the previous example. Empty return should still work:
image

coledie: Interesting, this only works if there is also a yield ni the function because normally for i in None throws an error

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: Yes but we cant take advantage of the benefits of the csv reader with the different length rows. This looks like string.split with extra overhead?

mortid0: 1. What overhead are you talking about?
2. Add a test case with a node binary feature equal to string: "parsing, not splitting". I don't think existing parser can handle it.

src/python/deepgnn/graph_engine/data/citation.py
mortid0: Please add a test to the converter.

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: 1. I benchmarked, CSV reader is faster even for just string splitting
2. I dont think either soltion will support this, I'll parameterize the split operators

coledie: Actually I forgot to do next(csv.reader), it is actually 25-50% slower than .split

docs/graph_engine/data_spec.rst
```diff
@@ -12,14 +12,86 @@ DeepGNN supports both homogeneous and heterogeneous graphs. Nodes and Edges supp
...
+	<node info>
+	<edge_1_info>
+	<edge_2_info>
+	...
+
+node_info: node_id,-1,node_type,node_weight,<features>
+edge_info: src,dst,edge_type,edge_weight,<features>
+features[dense]: dtype_name,length,v1,v2,...,dtype_name2,length2,v1,v2,...
+features[sparse with 2 dim coordinates vector]: dtype_name,values.size/coords.shape[1],c1,c2,...,v1,v2,...
+features[sparse with 1 dim coordinates vector]: dtype_name,values.size/0,c1,c2,...,v1,v2,...
```
mortid0, line 46: what does `/0` mean?

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,199 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+        self,
+        default_node_type: Optional[int] = None,
+        default_edge_type: Optional[int] = None,
+        default_node_weight: Optional[float] = None,
+        default_edge_weight: Optional[float] = None,
+        default_node_feature_types: Optional[List[Optional[str]]] = None,
+        default_edge_feature_types: Optional[List[Optional[str]]] = None,
+        default_node_feature_lens: Optional[List[Optional[List[int]]]] = None,
+        default_edge_feature_lens: Optional[List[Optional[List[int]]]] = None,
+        delimiter=",",
```
mortid0, line 124: where `delimiter` is documented?

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,199 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+        length_delimiter="/",
```
mortid0, line 125: how do you escape these delimiters for string features?

coledie: Its with a backslash, Its at the bottom of the docstring.

docs/graph_engine/data_spec.rst
coledie: It is a sparse feature where the coordinates vector is only one dimensional.

mortid0: Could you add this to the doc?

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -720,6 +775,255 @@ def test_sanity_edge_sparse_features_data(graph_with_sparse_features):
...
+        gen = decoder.decode("0,1,4")
+        next(gen)
+    gen = decoder.decode("0,1,4,1")
+    next(gen)
+
+
+def test_edge_list_binary_escape():
+    decoder = EdgeListDecoder()
+    src, dst, typ, weight, features = next(
+        decoder.decode("0,-1,0,1.0,binary,1,test\,feature")
```
mortid0, line 997: Can you add a few more test cases:
* "0,-1,0,1.0,binary,1,test\,feature"
* "0,-1,0,1.0,binary,1,test\\,feature"
* "0,-1,0,1.0,binary,1,test\\\,feature"
* "0,-1,0,1.0,binary,1,test\,feature\"
* "0,-1,0,1.0,binary,1,test\,feature\,"
* "0,-1,0,1.0,binary,1,test\,feature,"
* "0,-1,0,1.0,binary,1,test\,\,feature"
* "0,-1,0,1.0,binary,1,\test,\\feature"
* "0,-1,0,1.0,binary,1,\,feature"

mortid0: Sorry, there are more test cases, md has it's own escape rules..

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -720,6 +775,303 @@ def test_sanity_edge_sparse_features_data(graph_with_sparse_features):
...
+    )
+    assert features[0] == "test"
+    src, dst, typ, weight, features = next(
+        decoder.decode("0,-1,0,1.0,binary,1,test\,feature")
+    )
+    assert features[0] == "test,feature"
+    src, dst, typ, weight, features = next(
+        decoder.decode("0,-1,0,1.0,binary,1,test\\,feature")
+    )
+    assert features[0] == "test,feature"
```
mortid0, line 1007: Why the feature value not `test\`?

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -720,6 +775,303 @@ def test_sanity_edge_sparse_features_data(graph_with_sparse_features):
...
+    )
+    assert features[0] == "test\,feature"
+    with pytest.raises(ValueError):
+        src, dst, typ, weight, features = next(
+            decoder.decode("0,-1,0,1.0,binary,1,test,feature,")
+        )
+    src, dst, typ, weight, features = next(
+        decoder.decode("0,-1,0,1.0,binary,1,test\,feature,")
+    )
+    assert features[0] == "test,feature"
```
mortid0, line 1019: what is the value of features[1]?

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -720,6 +775,303 @@ def test_sanity_edge_sparse_features_data(graph_with_sparse_features):
...
+    src, dst, typ, weight, features = next(
+        decoder.decode("0,-1,0,1.0,binary,1,test\,feature\\")
+    )
+    assert features[0] == "test,feature"
```
mortid0, line 1023: not `test,feature\`?

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -720,6 +775,309 @@ def test_sanity_edge_sparse_features_data(graph_with_sparse_features):
...
+    )
+    assert features[0] == "test"
+    src, dst, typ, weight, features = next(
+        decoder.decode("0,-1,0,1.0,binary,1,test\,feature")
+    )
+    assert features[0] == "test,feature"
+    src, dst, typ, weight, features = next(
+        decoder.decode(r"0,-1,0,1.0,binary,1,test\\,feature")
+    )
+    assert features[0] == r"test\,feature"
```
mortid0, line 1007: This is a wrong result, there should be 2 features: `test\` and `feature`

coledie: In order to get 2 features it would have ot be "binary,1,test,binary,1,feature", maybe it should error out instead of silently ignoring?

mortid0: Looks like at least 3 more test cases are needed:
1. `0,-1,0,1.0,binary,1,test\\,binary,1,hello` -> 2 features
2. `0,-1,0,1.0,binary,2,test\\,hello` -> 2 features
3. `0,-1,0,1.0,binary,1,test\\,hello` -> error

coledie: Second test is invalid, each feature vector needs a key and length. Binary_feature only supports length 1. I'll add the other two.

mortid0: >Second test is invalid, each feature vector needs a key and length. 

You still need to test a combination of multiple binary features:
* `0,-1,0,1.0,binary,1,test\,hello,binary,1,test`
* `0,-1,0,1.0,binary,1,test\,hello\\,binary,1,test`
and so on..

mortid0:
why target branch is not main?

coledie:
It requires the node / edge on different partitions support.

mortid0:
That PR was approved a week ago, is there anything missing there?

src/cc/lib/graph/graph.cc
```diff
@@ -99,8 +99,14 @@ void Graph::GetNodeType(std::span<const NodeId> node_ids, std::span<Type> output
...
         else
         {
-            const auto index = internal_id->second;
-            *curr_type = m_partitions[m_partitions_indices[index]].GetNodeType(m_internal_indices[index]);
+            auto index = internal_id->second;
+            size_t partition_count = m_counts[index];
+            for (size_t partition = 0; partition < partition_count; ++partition, ++index)
+            {
+                *curr_type = m_partitions[m_partitions_indices[index]].GetNodeType(m_internal_indices[index]);
+                if (*curr_type != snark::DEFAULT_NODE_TYPE)
```
mortid0, line 107: Why there are 2 default types here? One is passed as an argument, and another is a constant. Rename to `[missing|stub|...]_node_type`?

mortid0, line 107: how do you handle a case when a node type is present only in one partition as `snark::DEFAULT_NODE_TYPE`? Where do we assign `default_type` to returned elements?

docs/graph_engine/data_spec.rst
```diff
@@ -4,22 +4,100 @@ Preparing Graph Data
 
 DeepGNN supports both homogeneous and heterogeneous graphs. Nodes and Edges support the following attributes,
 
-  * Node/Edge Type: `int`.
+  * Node/Edge Type: `int, >= 0`.
```
mortid0, line 7: Why do we still use it in [local.py](https://github.com/microsoft/DeepGNN/blob/cdieckhaus/externalize_linear/src/python/deepgnn/graph_engine/snark/local.py#L197)?

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
```diff
@@ -1015,25 +1132,87 @@ def test_node_sampling_distributed_graph_multiple_partitions_raises_empty_types(
...
+def test_distributed_graph_node_types(multi_partition_graph_data, storage_type):
+    address = [f"localhost:{find_free_port()}", f"localhost:{find_free_port()}"]
+    s1 = server.Server(
+        multi_partition_graph_data, [0], hostname=address[0], storage_type=storage_type
+    )
+    s2 = server.Server(
+        multi_partition_graph_data, [1], hostname=address[1], storage_type=storage_type
+    )
+    cl = client.DistributedGraph(address)
+    types = cl.node_types(np.array([9, 5, 0], dtype=np.int64), -1)
```
mortid0, line 1193: why do you use default type -1? Types should be only positive according to `data_spec.rst`.

src/cc/lib/graph/partition.cc
```diff
@@ -283,9 +283,18 @@ Type Partition::GetNodeType(uint64_t internal_node_id) const
...
+
+bool Partition::HasNodeFeatures(uint64_t internal_node_id) const
+{
+    return GetNodeType(internal_node_id) != snark::DEFAULT_NODE_TYPE;
+}
+
+bool Partition::GetNodeFeature(uint64_t internal_id, std::span<snark::FeatureMeta> features,
                                std::span<uint8_t> output) const
 {
+    if (!HasNodeFeatures(internal_id))
```
mortid0, line 295: Why do we have to check feature presence by the node type instead of feature index? On line 307 we fill `output` with 0s, can we return false from the method and fill zeros in the client? Also less encoding/decoding in protobufs..

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -720,6 +775,374 @@ def test_sanity_edge_sparse_features_data(graph_with_sparse_features):
...
+    with pytest.raises(RuntimeError):
+        src, dst, typ, weight, features = next(
+            decoder.decode("0,-1,0,1.0,binary,1,test\,feature,binary,1,feature,")
+        )
+    src, dst, typ, weight, features = next(
+        decoder.decode("0,-1,0,1.0,binary,1,test\,feature\,,binary,1,feature")
+    )
+    assert features[0] == "test,feature,"
+    assert features[1] == "feature"
+    src, dst, typ, weight, features = next(
```
mortid0, line 1118: I don't think this test case will be handled correctly:
```python
src, dst, typ, weight, features = next(
        decoder.decode(r"0,-1,0,1.0,binary,1,test\,\,feature\,,binary,1,\\")
    )
assert features[1] == "\\"
```

docs/graph_engine/data_spec.rst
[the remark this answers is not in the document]
coledie: This is not new behavior, it was previously undocumented. It would have errored if someone added.

coledie: No actual node will have type = -1, if some node in the graph has type "-1" it should have no features it is only for edge sorting.

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
coledie: A real node cannot have type == -1, so this is a good return to signify null value.

src/cc/lib/graph/graph.cc
coledie: If the node only exists in one partition with the default type, this means the node does not exist and nothing should be returned. The default type is handled in python I believe.

coledie: Change to placeholder_node_type since nodes with this type are placeholder for edge sorting.

src/python/deepgnn/graph_engine/data/citation.py
```diff
@@ -264,48 +263,34 @@ def _write_json_graph(
...
                     label=int(labels[nid]),
                     neighbors=adj[nid].nonzero()[1],
                 )
                 fout.write(tmp)
-                fout.write("\n")
 
     NODE_TYPE_ID = {"train": 0, "val": 1, "test": 2, "other": 3}
 
-    def to_json_node(
+    def to_edge_list_node(
```
mortid0, line 277: why this change is needed?

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,217 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+                key, length = None, None  # type: ignore
+
+            if key is None:
+                try:
+                    key = next(data)
+                except StopIteration:
+                    break
+                try:
+                    int(key)
+                    raise ValueError("Expected feature vector key is str.")
```
mortid0, line 246: Typo: you mean key is supposed to be int, not str, because of line 245? Where do we use the result of the conversion to int?

src/python/deepgnn/graph_engine/data/citation.py
coledie: It makes the data loaders about twice as fast.

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: We are trying to check if things like "1" is a string holding an integer, we want it to error out.

src/python/deepgnn/graph_engine/data/citation.py
mortid0: This is a public method, pease update the changelog.

src/python/deepgnn/graph_engine/snark/decoders.py
mortid0: Why do you check it here? You are still passing a string to `self._get_feature`. It is very confusing order of operations.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Add new converter input format "EdgeList" with EdgeListDecoder. Format has nodes and edges on separate lines, is smaller and faster to convert.
+
+- Rename function deepgnn.graph_engine.data.to_json_node -> deepgnn.graph_engine.data.to_edge_list_node and update functionality accordingly.
```
mortid0, line 12: This should be in `Changed` section.

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,219 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+
+    def _get_feature(
+        self, key: str, length: List[int], data: Iterator
+    ) -> Union[np.ndarray, Tuple[np.ndarray, np.ndarray], None]:
+        def _str_get(item, idx):
+            if not len(item):
+                return ""
+            return item[idx]
+
+        try:
```
mortid0, line 165: I still don't understand what this code does:
1. Why do you want to convert to int and immediately throw an exception? and later use key as a `dtype` to create arrays. Both strings `binary` and `some other random string` will raise `ValueError` if you'll try to convert it to int.
2. Where do you check `Expected feature vector key is str.` message in tests? If the error is silently swallowed inside the code, why do we need it?

docs/graph_engine/data_spec.rst
```diff
@@ -12,14 +12,93 @@ DeepGNN supports both homogeneous and heterogeneous graphs. Nodes and Edges supp
...
+	<edge_1_info>
+	<edge_2_info>
+	...
+
+node_info: node_id,-1,node_type,node_weight,<features>
+edge_info: src,dst,edge_type,edge_weight,<features>
+features[dense]: dtype_name,length,v1,v2,...,dtype_name2,length2,v1,v2,...
+features[sparse with 2 dim coordinates vector]: dtype_name,values.size/coords.shape[1],c1,c2,...,v1,v2,...
+features[sparse with 1 dim coordinates vector]: dtype_name,values.size/0,c1,c2,...,v1,v2,...
+* Sort file as node, then all of the nodes outgoing edges sorted by type then dst.
```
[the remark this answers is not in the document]
coledie: Lexographic sorting is problematic for the edge ordering. Edges must be sorted first by type then dst -- with dst being sorted in same order as nodes. Users can use ETL jobs to sort this file format which makes it much easier.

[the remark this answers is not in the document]
mortid0: > Lexographic sorting is problematic for the edge ordering.

Please, explain, why is problematic, why can't we change columns and make everything more natural?

> Users can use ETL jobs to sort this file format which makes it much easier.

We don't need any ETL pipelines, just a command line utility `sort` to work with lexicographic ordering: it implements an external sort algorithm and can process almost any file.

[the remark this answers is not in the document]
coledie: Blocker is edge sort, edges need to be sorted by dst in increasing order of internal id, so if the user keeps track of this ordering the linear format will work for that scenario.

src/cc/lib/graph/partition.cc
mortid0: We don't need node_type to distinguish that node doesn't have any features. The problem with current approach is it doesn't allow a node to have features on different partitions. To resolve it we need to merge on the client, instead of filling zeros on the server. However, it is not related to this PR and will be fixed later with temporal feature.

docs/graph_engine/data_spec.rst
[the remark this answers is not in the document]
coledie: Why would this make it work?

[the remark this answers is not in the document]
mortid0: Because you can sort edges/nodes in a lexicographic order before starting conversion.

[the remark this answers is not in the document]
coledie: Why does reordering the columns enable lexicographic sorting?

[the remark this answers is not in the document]
mortid0: Current implementation:
```
# src, dst, type, data
1,3,2,
1,-1,1,
1,2,2,
```
Is hard to sort in the proposed format. Changed order before sorting:
```
# src, type, dst, data
1,2,3,
1,1,-1,
1,2,2,
```
after:
```
# src, type, dst, data
1,1,-1
1,2,2,
1,2,3,
```

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -722,6 +776,378 @@ def test_sanity_edge_sparse_features_data(graph_with_sparse_features):
...
+def test_edge_list_binary_escape():
+    decoder = EdgeListDecoder()
+    src, dst, typ, weight, features = next(decoder.decode("0,-1,0,1.0,binary,1,test"))
+    assert features[0] == "test"
+    with pytest.raises(RuntimeError):
+        src, dst, typ, weight, features = next(
+            decoder.decode("0,-1,0,1.0,binary,1,test,feature")
+        )
+    src, dst, typ, weight, features = next(
+        decoder.decode("0,-1,0,1.0,binary,1,test\,feature")
```
mortid0, line 1005: Please, make sure that it is ok to use spaces after comas:
```python
    src, dst, typ, weight, features = next(
        decoder.decode("0, -1, 0, 1.0, binary, 1,test\,feature")
    )
```

coledie: Users can enable this by overriding the delimiter=", " which is tested.

mortid0: ", " is not a proper delimiter, because parsers typically skip spaces.

docs/graph_engine/data_spec.rst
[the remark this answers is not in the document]
coledie: This works: sort edgelist.csv -t, -n -k1,1 -k3,3 -k2,2 --parallel=1
- You have to specify the column order to sort anyways for a csv so we don't need to actually switch them. I added to docs. But this does require you to invert the type of the node.
-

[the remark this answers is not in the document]
mortid0: Lexicographical order is way faster:
```sh
time sort ~/tmp/edges.txt -t, -n > ~/tmp/lexi.txt

real    1m46.391s
user    6m53.227s
sys     0m7.406s
```
vs
```sh
time sort ~/tmp/edges.txt -t, -n -k1,1 -k3,3 -k2,2 > ~/tmp/custom.txt

real    2m46.870s
user    9m47.953s
sys     0m8.420s
```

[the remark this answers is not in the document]
coledie: lexographical sort does not work for non trivial files:
1,3,0,
1,-1,-1,
1,3,11,
1,2,11,
1,1,11,
1,0,11,
1,2,2,
2,-1,-5,
2,1,1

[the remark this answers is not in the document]
mortid0: Set env variable `LC_COLLATE=C` or `LC_ALL=C` to sort negative numbers properly.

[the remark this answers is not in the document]
coledie: I'm saying if you run lexographical sort on that file, it will not work, it acts like each line is just one big number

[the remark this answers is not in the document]
mortid0: Hmm, ok. So why can't we use a natural column order: `1, 2, 3` and have to use `1, 3, 2`?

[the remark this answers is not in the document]
coledie: src and dst are the only columns that are strictly required, the rest can be defaulted and skipped. So it makes sense that src and dst goes first and everything else after.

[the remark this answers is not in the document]
mortid0: We are talking about positions, not field names.
* If users have a default edge type, then sorting is based on columns `1, 2` (src and dst).
* If there are more than one type, then types should go before dst and sorting should be `1, 2, 3`, not `1, 3, 2`.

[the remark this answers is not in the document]
coledie: Every sort function makes it easy to choose which columns have priority. It is confusing to everyone when the column order has to change like that, instead we keep it extensible
src, dst
src, dst, type, weight, features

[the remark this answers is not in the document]
mortid0: > Every sort function makes it easy to choose which columns have priority. 

It is easy to forget that edges should be sorted by type first. We already made [such mistake](https://github.com/microsoft/DeepGNN/commit/dd9c7c32ff9209d829004ad17242e37929fa70da) and format should help us to avoid it.

> It is confusing to everyone when the column order has to change like that, instead we keep it extensible. 

Not at all, the actual format is:
`src, type[optional], dst, weight[optional], features[optional]`.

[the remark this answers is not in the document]
coledie: We can add an assertion to prevent confusion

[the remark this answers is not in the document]
mortid0: 1. We are talking about data format, not code. You can't add assertion to data.
2. This thread is growing way to long and so far you didn't show any benefits of changing a natural column order. `src`/`type`/`destination` is kind of how edges are stored internally, so it will be always easier to keep input in a similar format.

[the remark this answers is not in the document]
coledie: 1. We already have an assertion in linear decoder to warn the user if they do this, we need the assertion either way to ensure their data is sorted.
2. This format makes sense
```
src, dst, type[optional], weight[optional], features[optional]
```
It does not make sense to have an optional variable in the middle of required variables, this is confusing
```
src, type[optional], dst, weight[optional], features[optional]
```
3. We already have the decoder in src,dst,type format, it will be easier to keep this input in the same format.
4. Edge  binaries are written in src,dst,type format, it will be easier to keep this input in the same format: https://github.com/microsoft/DeepGNN/blob/cdieckhaus/externalize_linear/src/python/deepgnn/graph_engine/snark/converter/writers.py#:~:text=%22%3DQQif%22%2C,weight%2C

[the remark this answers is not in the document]
mortid0: 1. I am talking about data format, not code. Why are you talking about assertions?
2. It seems you are using python style for passing parameters. Optional arguments are totaly fine anywhere in the message. See [protocol-buffers for example](https://developers.google.com/protocol-buffers/docs/overview#syntax)
3. Decoder is actually broken, code is easier to fix than data. Here is an example. If we want to do a quick conversion with just a binary writer then we need to use something like this with `[src, dst, type, weight, feature]` order:
```python
from deepgnn.graph_engine.snark.converter.writers import BinaryWriter
bw = BinaryWriter(".", 0, True, True)
edges = []
with open("graph.txt", "r") as f:
    lines = f.readlines()
    raw_edges = [[int(x) for x  in line.split(",")] for line in lines]
    edges = list(map(lambda x: (x[0], x[2], x[1], 1, []), raw_edges))
dst_sorted = sorted(edges, key=lambda x: x[1])
tp_sorted = sorted(dst_sorted, key=lambda x: x[2])
src_sorted = sorted(tp_sorted, key=lambda x: x[0])
bw.add(src_sorted)
bw.close()
```
However, if we move type to a second place `[src, type, dst, weight, feature]` code will be much shorter:
```python
from deepgnn.graph_engine.snark.converter.writers import BinaryWriter
bw = BinaryWriter(".", 0, True, True)
edges = []
with open("graph.txt", "r") as f:
    lines = f.readlines()
    raw_edges = [[int(x) for x  in line.split(",")] for line in lines]
    edges = list(map(lambda x: (x[0], x[2], x[1], 1, []), raw_edges))
bw.add(sorted(edges))
bw.close()
```
4. If you read the comments to that code, you'll see that we were trying to avoid using extra memory. Graph engine actually transforms that data to [CSR format](https://github.com/microsoft/DeepGNN/blob/main/src/cc/lib/graph/partition.h#L118) where we store sorted types first and then all destinations. I don't know yet how to convert directly to CSR format on the fly yet and that's why we have temporary files with both dst and tp.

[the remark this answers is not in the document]
coledie: #2 I'm saying it's confusing to explain to people, as opposed to saying just add an extra comma you have to tell them to put it between the two. For the sort in #3, its actually sorted(edges, key=lambda x: (x[0], x[2], x[1])).
But I actually tried to implement this and its not possible // easy. We need to know if dst == -1 or not, if the given line is a node or not, in order to determine whether or not there is a default value for the type given // whether dst is in column 2 or 3.

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -722,6 +776,386 @@ def test_sanity_edge_sparse_features_data(graph_with_sparse_features):
...
+        assert len(result) == expected_size
+        assert result[0:8] == (0).to_bytes(8, byteorder=sys.byteorder)
+        assert result[8:16] == (0).to_bytes(8, byteorder=sys.byteorder)
+        assert result[16:20] == (0).to_bytes(4, byteorder=sys.byteorder)
+        assert result[20:28] == (2).to_bytes(8, byteorder=sys.byteorder)
+        assert result[28:36] == (1).to_bytes(8, byteorder=sys.byteorder)
+        assert result[36:40] == (0).to_bytes(4, byteorder=sys.byteorder)
+
+
+def test_edge_list_error_checking():
```
mortid0, line 961: where are tests for sparse feature parsing?

docs/graph_engine/data_spec.rst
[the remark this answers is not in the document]
mortid0: why can't we use type=-1 to encode a node item?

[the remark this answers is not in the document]
coledie: I'll do this, it will also mean the sort wont require you to invert node types

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
coledie: All of the normal ones, its in the pytest.parameterize: https://github.com/microsoft/DeepGNN/blob/cdieckhaus/externalize_linear/src/python/deepgnn/graph_engine/snark/tests/convert_test.py#:~:text=def%20graph_with_sparse_features(request)%3A,output%20%3D%20tempfile.TemporaryDirectory()

mortid0: These are not parsing tests, but functional tests to verify a correct conversion logic across formats.
For parsing tests you need to handle various inputs for EdgeList only format, similarly what you did for binary features.

coledie: Why do we only need this for EdgeList format? EdgeList has all of the same sparse feature tests as Json and TSV.

mortid0: Because json format uses standard library to parse json strings, but EdgeList format is using a hand-made parser. TSV doesn't support sparse features as I recall.

coledie: I've added these.

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,225 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+
+        Sort the file so the first line has the first node's info, the next few lines have all the first node's
+        outgoing edges. Then the next line will have the second node's info and so on.
+
+        Feature fectors to fill <features> can be dense or sparse. Features will be given
+        indexes starting at 0 and indexes can be skipped with 0 length vectors. Each node and
+        edge does not have to have the same number of features or feature types.
+
+        features[dense]: dtype_name,length,v1,v2,...,dtype_name2,length2,v1,v2,...
+        features[sparse with 2 dim coordinates vector]: dtype_name,values.size/coords.shape[1],c1,c2,...,v1,v2,...
```
mortid0, line 69:
```python        
        features[sparse with 2 dim coordinates vector]: dtype_name,values.size/coords.shape[1],c1,c2,...,v1,v2,...
        features[sparse with 1 dim coordinates vector]: dtype_name,values.size/0,c1,c2,...,v1,v2,...        
```

1. It is not clear what coords.shape[1] is and why do we use second dimension, not first.
2. Why there is a difference between 2 and 1 dimensional vectors? Can we just keep the same number?
3. How do we differentiate between multidimensional coordinate items? Does `c1,c2,..` mean coordinates for values `v1, v2` or coordinates for value `v1` for the first 2 dimensions?

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -43,6 +43,225 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
...
+        0,1,0,4,10,1,1,1
+        1,-1,1,1,1,1.1,1.1
+        1,0,0,4,10,1,1,1
+        ```
+
+    Delimiters
+        "," is the default column delimiter, it can be overriden with the delimiter parameter.
+        "/" is the default sparse features length delimiter, it can be overriden with the
+            length_delimiter parameter.
+        "\" is the escape for the delimiter in "binary" features.
```
mortid0, line 121: 1. why first two delimiters can be overridden, but not third one?
2. what will happen if some of them are the same, e.g. sparse features length delimiter is "\"?

src/python/deepgnn/graph_engine/snark/decoders.py
coledie: I've updated these.

mortid0 (approved):

PR #60, opened 2022-07-27
Fix dimensions in sparse features

mortid0:

Previous Behavior
----------------
Dimensions of sparse features were not filled with 0s, so if some features were missing for nodes in a partition, we overwrite it with 0 and return garbled results, because dimensions are used later to parse feature data.

New Behavior
----------------
Initialize dimensions with 0s in memory and distributed graph.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.1 DeepGNN
 
+## 0.1.52 - 2022-07-27
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 0.1 DeepGNN
 
+## 0.1.52 - 2022-07-27
```
coledie, line 5: Should be ### ? So it is on different spacing.

coledie (approved):

PR #72, opened 2022-08-11
GIN Model

adithyabhonsley:
- [ ] Forked repo is synced with upstream -> github shows no code delta outside of the desired.
- [ ] Documentation is added or updated to reflect new code in the same format as the rest of the repo.

Previous Behavior
----------------
* Provide relevant issue number if applicable, eg #44.

New Behavior
----------------

* Graph Isomorphic Network (GIN) model for training large-scale graphs


examples/pytorch/gin/model.py
```diff
@@ -0,0 +1,258 @@
...
+        # Linear Function: hidden rep -> prediction score
+        self.linears_prediction = torch.nn.ModuleList()
+        for layer in range(num_layers):
+            if layer == 0:
+                self.linears_prediction.append(nn.Linear(input_dim, output_dim))
+            else:
+                self.linears_prediction.append(nn.Linear(hidden_dim, output_dim))
+    
+    def next_layer(self, h, layer, features, nb_counts, num_nodes):
+        pooled_rep = self.mlps[layer](h)
```
coledie, line 141: Looks like H is not pooled here like in the example, this is probably important for learning

examples/pytorch/gin/model.py
```diff
@@ -0,0 +1,258 @@
...
+        for layer in range(self.num_layers - 1):
+            h = self.next_layer(h, layer, features, nb_counts, num_nodes)
+            hidden_rep.append(h)
+
+        score = 0
+        
+        for layer, h in enumerate(hidden_rep):
+            # pooled_h = torch.spmm(nb_features, h)
+            # get_logger().info("Layer " + str(layer) + " | h shape: " + str(pooled_h.shape))
+            score += F.dropout(self.linears_prediction[layer](h), self.final_dropout, training = self.training)
```
coledie, line 185: You might be able to use an Aggregator to handle the pooling here "        ###create sum or average pooling sparse matrix over entire nodes in each graph (num graphs x num nodes)
" 
I'm pretty sure pooling is important for learning

examples/pytorch/gin/model.py
```diff
@@ -0,0 +1,300 @@
...
+        #     strategy = "randomwithoutreplacement"
+        # )[0]
+
+        # get_logger().info("LEN 1: " + str(len(context['neighbors'])))
+        # get_logger().info("**********************************************************")
+
+        # get_logger().info("USING NORMAL NEIGBORS *****************************")
+        context["neighbors"] = graph.neighbors(
+            context["inputs"], np.array(self.edge_type)
+        )[0][: len(context["inputs"])]
```
coledie, line 268: I think this slice is not right, I think you are dropping a lot of neighbors? If you just want say 10 neighbors per node you should use uniform sample neighobrs instead of this which is full neighbors.

mortid0 (changes requested):
Looks like there are lots of merge conflict changes and binary files. Could you clean up the PR first?

PR #75, opened 2022-08-18
Fully add python typing.

coledie:

New Behavior
----------------
* Examples folder is now type checked to help verify API is accurate.
* Add and updates types on all python/deepgnn folders: graph_engine, torch and tf.
* Refactor some functions.
* Query output type and forward input type is now QueryOutput instead of dict.


.github/actions/lint/action.yml
```diff
@@ -12,7 +12,7 @@ runs:
     - run: brew install clang-format
       shell: bash
       if: runner.os == 'macOS'
-    - run: pip install wheel pre-commit==2.17.0 mypy==0.780 numpy==1.21.6
+    - run: pip install wheel pre-commit==2.17.0 mypy==0.971 numpy==1.21.6 torch==1.11.0 tensorflow==2.7.2
```
mortid0, line 15: why do we need to install tensorflow/torch for a linter?

.mypy.ini
```diff
@@ -3,7 +3,7 @@ warn_return_any = False
 warn_unused_configs = True
 mypy_path = src/python/deepgnn
 
-files=src/python/deepgnn
+files=src/python/deepgnn,examples/*/*/*.py
```
mortid0, line 6: nit: use globbing to much every file in the repo `**/*.py`

examples/pytorch/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""DeepGNN model examples."""
```
mortid0, line 3: nit: """DeepGNN Pytorch model examples."""

examples/pytorch/gat/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""DeepGNN model examples."""
```
mortid0, line 3: nit: """DeepGNN Pytorch GAT model."""

examples/pytorch/geometric/gat/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""DeepGNN model examples."""
```
mortid0, line 3: nit: """DeepGNN GAT model implemented with Pytorch Geometric."""

mortid0: same in other init files.

examples/pytorch/graphsage/main.py
```diff
@@ -19,7 +19,7 @@
     SamplingStrategy,
     GraphEngineBackend,
 )
-from model import SupervisedGraphSage, UnSupervisedGraphSage
+from model import SupervisedGraphSage, UnSupervisedGraphSage  # type: ignore
```
mortid0, line 22: why do we need to ignore types in imports?

.github/actions/lint/action.yml
coledie: Otherwise it will ignore torch.Tensor and tf.Tensor types - it does not add any time to CI pre-commit it looks like.

examples/pytorch/graphsage/main.py
coledie: Mypy does not handle the repeated filenames and it is hard to get the imports right for each of the examples. The most important thing is how the examples use the trainer and GE to verify those types, not so much the model implementation is typed right.

.mypy.ini
coledie: I've updated it to "src/python/deepgnn,examples" instead, globbing is hard since bazel need to be ignored.

.github/actions/lint/action.yml
mortid0: pip install runs for 1 minute instead of 5 seconds with python 3.9 on ubuntu. Can you [cache it with a requirements.txt file](https://github.com/actions/cache/blob/main/examples.md#multiple-oss-in-a-workflow-with-a-matrix) and reference it in the doc?

.mypy.ini
```diff
@@ -32,6 +32,9 @@ ignore_missing_imports = True
 [mypy-tensorflow.*]
 ignore_missing_imports = True
```
mortid0, line 33: Since we install tf/torch now, their missing imports should be removed, right?

mortid0 (approved):

.github/actions/lint/action.yml
coledie: The downloads are cached: https://github.com/microsoft/DeepGNN/runs/7928528362?check_suite_focus=true#:~:text=Collecting%20torch%3D%3D1.11.0,67

.mypy.ini
mortid0: use [exclude section](https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude)

coledie: I spent quite a while trying to get that to work and it seems to have weird behavior and I cannot get it to work. I think the order of the commands matter and maybe there are some overrides between commands or something that is not documented.

mortid0: Create an issue on the mypy github page and describe what you tried to do?

coledie: Right now files=src/python/deepgnn,examples which is not so bad. It is probably not worth the effort?

.mypy.ini
coledie: It works on my local, but not on CI. I think this is a separate issue: https://github.com/microsoft/DeepGNN/runs/7963445411?check_suite_focus=true#:~:text=examples/pytorch/link_prediction/poolers.py%3A5%3A%20error%3A%20Cannot%20find%20implementation%20or%20library%20stub%20for%20module%20named%20%22torch%22

.mypy.ini
mortid0: Well, you forgot to add `docs` folder, so it seems worth to use a correct pattern and exclude only bazel-bin folder.

.mypy.ini
mortid0: Does your local python version match CI version?

coledie: Yes and same package versions

.mypy.ini
coledie: docs does not have any code to type check. I dont think we should ever have code outside of src and examples to need a complex pattern.

PR #77, opened 2022-08-19
Add generator to automatically pull e2e test ports.

coledie:
- [X] Forked repo is synced with upstream -> github shows no code delta outside of the desired.

Previous Behavior
----------------
* Every server that runs in tests needs a different port to be manually configured.

New Behavior
----------------
* Add helper function to find valid ports.


src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
```diff
@@ -26,6 +26,17 @@
 import deepgnn.graph_engine.snark._lib as lib
 
 
+import socket
```
mortid0, line 29: nit: push these imports up, with the [rest of the standard library](https://peps.python.org/pep-0008/#imports).

mortid0 (approved):
Surprised this works on windows!

PR #79, opened 2022-08-23
Refactor distributed feature fetching

mortid0:

Previous Behavior
----------------
* Duplicated code for sparse and string feature extraction for nodes and edges in distributed client.

New Behavior
----------------
* Templated extraction code to work for both node and edge feature extraction.
* Log a fatal error if sparse features dimensions don't match between nodes on different shards.


coledie (approved):
Why did you do this refactor?

mortid0:
> Why did you do this refactor?

Need it for temporal features and I thought there might be a bug in how we handle sparse/string features, but apparently everything is ok.

PR #80, opened 2022-08-26
Introduce version for binary data.

mortid0:

Previous Behavior
----------------
* We have to communicate breaking changes in binaries directly.

New Behavior
----------------
* Metadata files now contain a version of the converter binary files were generated with. GE reads this metadata and verifies it can work with these binaries.

### DIFF
--- a/src/cc/lib/graph/metadata.h
+++ b/src/cc/lib/graph/metadata.h
@@ -10,6 +10,9 @@
 
 namespace snark
 {
+
+const size_t MINIMUM_SUPPORTED_VERSION = 1;
+
 struct Metadata
 {
     Metadata() = default;
@@ -17,6 +20,7 @@ struct Metadata
     void Write(std::filesystem::path path) const;
 
     // Graph information.
+    size_t m_version;
     size_t m_node_count;
     size_t m_edge_count;
     size_t m_edge_type_count;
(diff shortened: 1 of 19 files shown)


src/cc/lib/graph/metadata.h
```diff
@@ -10,13 +10,17 @@
 
 namespace snark
 {
+
+const size_t MINIMUM_SUPPORTED_VERSION = 1;
```
coledie, line 14: Can we avoid duplicate magic numbers or otherwise link to eachother? Minimum_supported_version should always be the same as the version right, since there is no need to update the version otherwise?

src/python/deepgnn/graph_engine/snark/meta.py
```diff
@@ -9,6 +9,9 @@
 from deepgnn.graph_engine.snark._lib import _get_c_lib
 import platform
 
+"""Version of binary files produced by converters to communicate breaking changes requiring regeneration of binary files."""
+VERSION = "v1"
```
coledie, line 13: BINARY_VERSION?

coledie (approved):

src/cc/lib/graph/metadata.h
mortid0: * It is hard to use a single value, one is in C++ runtime value, another is in Python runtime.
* Changes don’t have to be breaking, some might be incremental.

Issue #83, opened 2022-09-09
AttributeError: Can't pickle local object 'CDLL.__init__.<locals>._FuncPtr'


nabihach:
- [ ] Issue is labeled using the label menu on the right side.

Environment
-----------
* Python version: (python -V) 3.7.7
* deepgnn-ge Version: (python -m pip show deepgnn-ge) 0.1.55.1
* deepgnn-torch Version: (python -m pip show deepgnn-torch) 0.1.55.1
* deepgnn-tf Version: (python -m pip show deepgnn-tf) not installed
* OS: (Windows, Linux, ...) Windows 10 Enterprise

Issue Details
-----------
* What you did - code sample or commands run

I installed deepgnn-torch via pip in a virtual environment. Then I cloned the deepgnn repository, cd-ed into the `examples/pytorch/gat/`  and then ran `bash run.sh`

* Expected behavior

I expected the training script to run without issues.

* Actual behavior

I see the error 
`  File "c:\users\myid\appdata\local\programs\python\python37\lib\multiprocessing\reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
AttributeError: Can't pickle local object 'CDLL.__init__.<locals>._FuncPtr'
`

Full stack trace:
```
$ bash run.sh
+ DEVICE=cpu
++ dirname run.sh
+ DIR_NAME=.
+ GRAPH=/tmp/cora
+ python -m deepgnn.graph_engine.data.citation --data_dir /tmp/cora
c:\users\myid\appdata\local\programs\python\python37\lib\runpy.py:125: RuntimeWarning: 'deepgnn.graph_engine.data.citation' found in sys.modules after import of package 'deepgnn.graph_engine.data', but prior to execution of 'deepgnn.graph_engine.data.citation'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
[2022-09-09 14:46:04,150] {convert.py:100} INFO - worker 0 try to generate partition: 0 - 1
[2022-09-09 14:46:04,151] {_adl_reader.py:124} INFO - [1,0] Input files: ['C:/Users/myid/AppData/Local/Temp/cora\\graph.json']
[2022-09-09 14:46:04,782] {dispatcher.py:143} INFO - record processed: 1000
[2022-09-09 14:46:05,257] {dispatcher.py:143} INFO - record processed: 2000
[...]

PR #85, opened 2022-09-12
Windows use 0 dataloader workers.

coledie:
- [ x] Forked repo is synced with upstream -> github shows no code delta outside of the desired.

Previous Behavior
----------------
#83 

Error with run.sh scripts using FileNodeSampler on windows caused by non-zero dataloader workers.

New Behavior
----------------
When using windows, 0 dataloader workers will be used by default to avoid fork.

### DIFF
--- a/src/python/deepgnn/pytorch/training/factory.py
+++ b/src/python/deepgnn/pytorch/training/factory.py
@@ -4,6 +4,7 @@
 """Functions to start distributed training."""
 
 import argparse
+import platform
 import torch
 from typing import Optional, Callable, List
 from deepgnn import TrainerType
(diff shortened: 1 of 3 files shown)


src/python/deepgnn/pytorch/training/factory.py
```diff
@@ -117,7 +118,7 @@ def run_dist(
 
     num_workers = (
         0
-        if issubclass(dataset.sampler_class, (GENodeSampler, GEEdgeSampler))
+        if issubclass(dataset.sampler_class, (GENodeSampler, GEEdgeSampler)) or platform.system() == "Windows"
```
mortid0, line 121: Do we need to run examples in CI on windows to trigger this behavior?

mortid0 (approved):

src/python/deepgnn/pytorch/training/factory.py
coledie: Yes the original bug has been here for all deepgnn package versions. Should we add powershell run files to all the examples or install something so we can use bash in windows?

PR #86, opened 2022-09-13
MutliWorkersConverter expose file iterator class.

coledie:

Previous Behavior
----------------
TextFileIterator is hardcoded to be used as MutliWorkersConverter input file iterator.

New Behavior
----------------
Add parameter so users can override, specifically for Avro input file iterator.

### DIFF
--- a/src/python/deepgnn/graph_engine/snark/convert.py
+++ b/src/python/deepgnn/graph_engine/snark/convert.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 """Conversion functions to internal binary format."""
-from typing import Optional, Type
+from typing import Optional
 import multiprocessing as mp
 import math
 from operator import add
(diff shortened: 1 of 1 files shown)


src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -36,6 +36,7 @@ def __init__(
         dispatcher: Dispatcher = None,
         skip_node_sampler: bool = False,
         skip_edge_sampler: bool = False,
+        iterator_class: Type[TextFileIterator] = TextFileIterator,
```
mortid0, line 39: Please, follow existing convention and use `None` as a default value.

src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -102,7 +104,7 @@ def convert(self):
         )
 
         d = self.dispatcher
-        dataset = TextFileIterator(
+        dataset = self.iterator_class(
```
mortid0, line 107: all iterator classes have to use the same arguments in the constructor? You probably want to pass an object, rather than a class on line 39.

mortid0 (changes requested):

mortid0 (approved):

PR #91, opened 2022-09-21
Add debug mode to MultiWorkersConverter.

coledie:

Previous Behavior
----------------
* Multiprocessing in MultiWorkersConverter crashes instead of showing error messages, so it is very hard to debug EdgeList and custom decoders.

New Behavior
----------------
* Add debug mode to MultiWorkersConverter, using worker_count=0 will now disable multiprocessing and show error messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Breaking. Added version checks for binary data. Requires to convert graph data or add v1 at the top of meta files.
 
+- Add Debug mode to MultiWorkersConverter, using worker_count=0 will now show error messages.
+
+### Changed
+- Rename function deepgnn.graph_engine.data.to_json_node -> deepgnn.graph_engine.data.to_edge_list_node and update functionality accordingly.
+
 ## [0.1.55] - 2022-08-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Breaking. Added version checks for binary data. Requires to convert graph data or add v1 at the top of meta files.
 
-- Add Debug mode to MultiWorkersConverter, using worker_count=0 will now show error messages.
-
-### Changed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Breaking. Added version checks for binary data. Requires to convert graph data or add v1 at the top of meta files.
 
+- Add debug mode to MultiWorkersConverter, using worker_count=0 will now disable multiprocessing and show error messages.
```
mortid0, line 12: How would anyone find about the debug mode? I don't see it in the code without reading PR description.

src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -118,17 +118,31 @@ def convert(self):
                 num_workers=self.worker_count,
             )
 
-        for _, data in enumerate(self.file_iterator):
-            for line in data:
-                d.dispatch(line)
-
-        d.join()
+        if self.worker_count >= 1:
```
mortid0, line 121: I think we debug by using threads instead of processes, so exceptions can propagate to the main thread.

mortid0 (changes requested):

src/python/deepgnn/graph_engine/snark/convert.py
coledie: use_threads is enabled by default and no errors are propogated.

src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -118,17 +118,31 @@ def convert(self):
...
+            d = self.dispatcher
```
mortid0, line 122: You don't need a dispatcher in a single thread. Directly use writer.

coledie: worker_count = 1 means that the TextFileIterator uses N workers but the dispatcher will actually use n_partitions workers so it does need the dispatcher.

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -121,6 +121,32 @@ def close(self):
         self.node_alias.close()
         self.edge_alias.close()
 
+        # For MultiWorkersConverter with 0 workers
```
mortid0, line 124: Please, remove it. Binary writers should not coordinate conversion process.

mortid0 (changes requested):

mortid0 (approved):

PR #97, opened 2022-10-11
[Temporal graph][2/4] Feature extraction

mortid0:

Previous Behavior
----------------
1. Only static features supported.
2. Feature extraction relied on using node types/edge flags to identify missing features.

New Behavior
----------------
1. Every node/edge has an optional parameter timestamp associated with it. If parameter is present, then we are going to do a binary search to find the latest feature. Otherwise return raw feature data.
2. Use feature timestamps buffer to identify missing features and fill with zeros if it is missing.

### DIFF
--- a/src/cc/lib/benchmark/partition_benchmark.cc
+++ b/src/cc/lib/benchmark/partition_benchmark.cc
@@ -11,13 +11,14 @@
 #include "src/cc/lib/graph/graph.h"
 #include "src/cc/lib/graph/xoroshiro.h"
 #include "src/cc/tests/mocks.h"
+
 #include <benchmark/benchmark.h>
+#include <boost/random/geometric_distribution.hpp>
 
 namespace
 {
 // Create a global(within benchmark code) client to share among threads.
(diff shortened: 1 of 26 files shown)


coledie (approved):

src/cc/lib/benchmark/partition_benchmark.cc
```diff
@@ -62,7 +62,7 @@ static void BM_NODE_FEATURES(benchmark::State &state, snark::PartitionStorageTyp
     for (auto _ : state)
     {
         const size_t batch_size = state.range(0);
-        g_client->GetNodeFeature(std::span(input_nodes).subspan(offset, batch_size), std::span(features),
+        g_client->GetNodeFeature(std::span(input_nodes).subspan(offset, batch_size), {}, std::span(features),
```
coledie, line 65: Did you check benchmarks remain the same after this change? Should a benchmark be added for timestamp filter?

mortid0:
Merging with https://github.com/microsoft/DeepGNN/pull/176

PR #101, opened 2022-10-21
1. Ray Dataset

coledie:

Previous Behavior
----------------
* We use custom DeepGNN dataset and sampler objects.

New Behavior
----------------
* Allow user to define ray data pipelines for usage in GE, https://docs.ray.io/en/latest/data/pipelining-compute.html#pipelining-datasets
* Add __reduce__ method to GE that will reinitialize graph engnine when it is de-serialized witihin ray data storage.
* Add delayed start parameter to MemoryGraph to delay GE load until after initialization.


mortid0 (commented):
Looks like there are more changes to come:
1. You removed GraphEngineBackend class, but it is used in [lots of places that haven't been updated yet in this PR](https://github.com/microsoft/DeepGNN/search?q=GraphEngineBackend).
2. If you are going to introduce ray to organize training process, then our shell scripts in examples should be replaced with regular python scripts.

I'd rather suggest to stack this PR in a following way:
1. Ray documentation
2. Replace one of the examples and benchmark ray trainer vs existing.
3. If step 2 looks good, replace existing trainers with ray trainers.

coledie:
Remake with local branch.

PR #106, opened 2022-10-26
Load each partition from separate folder.

### LINKED ISSUE #105: Add option to load each partition from its own folder

mortid0:

Previous Behavior
----------------
* Each partition has to be loaded from one folder.

New Behavior
----------------
* Use list of tuples for each partition (instead of integers before): path and id to specify where to load data from. Meta file can be located anywhere and is a parameter to GE as well.

Closes #105

### DIFF
--- a/docs/graph_engine/custom_decoder.rst
+++ b/docs/graph_engine/custom_decoder.rst
@@ -131,7 +131,7 @@ We load the generated binaries into a graph engine and demonstrate it working.
 
     >>> import deepgnn.graph_engine.snark.client as client
     >>> import numpy as np
-    >>> cl = client.MemoryGraph(working_dir.name, [0])
+    >>> cl = client.MemoryGraph(working_dir.name, [(working_dir.name, 0)])
     >>> cl.node_features(nodes=[0, 1], features=[[0, 2]], dtype=np.float32)
     array([[3.5, 6. ],
            [2. , 3. ]], dtype=float32)
--- a/docs/graph_engine/from_networkx.rst
+++ b/docs/graph_engine/from_networkx.rst
@@ -82,7 +82,7 @@ Create a client to use from the temp folder:
 
 .. code-block:: python
 
-	>>> cl = client.MemoryGraph(working_dir.name, [0])
+	>>> cl = client.MemoryGraph(working_dir.name)
 	>>> cl.node_features(nodes=[1], features=[[0, 2]], dtype=np.float32)
 	array([[1.        , 0.51667684]], dtype=float32)
 
--- a/src/cc/lib/graph/partition.cc
+++ b/src/cc/lib/graph/partition.cc
@@ -26,8 +26,9 @@ struct EdgeRecord
     float m_weight;
 };
 } // namespace
-Partition::Partition(std::filesystem::path path, std::string suffix, PartitionStorageType storage_type)
-    : m_metadata(path), m_storage_type(storage_type)
+Partition::Partition(Metadata metadata, std::filesystem::path path, std::string suffix,
+                     PartitionStorageType storage_type)
+    : m_metadata(std::move(metadata)), m_storage_type(storage_type)
 {
     ReadNodeMap(path, suffix);
     ReadNodeFeatures(path, suffix);
--- a/src/python/deepgnn/graph_engine/snark/client.py
+++ b/src/python/deepgnn/graph_engine/snark/client.py
@@ -22,7 +22,7 @@ from ctypes import (
     c_size_t,
     c_uint32,
 )
-from typing import Any, List, Tuple, Union, Optional
+from typing import Any, List, Tuple, Union, Optional, Sequence
 from enum import IntEnum
 
 import numpy as np
@@ -147,8 +147,8 @@ class MemoryGraph:
 
     def __init__(
         self,
-        path: str,
-        partitions: List[int] = [0],
+        meta_path: str,
+        partitions: Sequence[Union[int, Tuple[str, int]]] = None,
         storage_type: PartitionStorageType = PartitionStorageType.memory,
         config_path: str = "",
         stream: bool = False,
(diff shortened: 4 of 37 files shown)


coledie (changes requested):

docs/graph_engine/custom_decoder.rst
```diff
@@ -131,7 +131,7 @@ We load the generated binaries into a graph engine and demonstrate it working.
 
     >>> import deepgnn.graph_engine.snark.client as client
     >>> import numpy as np
-    >>> cl = client.MemoryGraph(working_dir.name, [0])
+    >>> cl = client.MemoryGraph(working_dir.name, [(working_dir.name, 0)])
```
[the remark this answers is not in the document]
mortid0: Of course, we do: one is for meta.txt and the other is for partition 0.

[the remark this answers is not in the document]
mortid0: I understand repeating dirs doesn't look good, but its simplicity relieves us from thinking about corner cases with a complex approach. What to do with empty integer lists: load the entire folder or nothing? How to treat repeating dirs and corresponding lists, merge and dedupe them?
Users can choose any python list comprehension technique they want to construct a final list.

src/cc/lib/graph/partition.cc
```diff
@@ -26,8 +26,9 @@ struct EdgeRecord
     float m_weight;
 };
 } // namespace
-Partition::Partition(std::filesystem::path path, std::string suffix, PartitionStorageType storage_type)
-    : m_metadata(path), m_storage_type(storage_type)
+Partition::Partition(Metadata metadata, std::filesystem::path path, std::string suffix,
+                     PartitionStorageType storage_type)
+    : m_metadata(std::move(metadata)), m_storage_type(storage_type)
```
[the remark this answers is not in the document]
mortid0: We use it to initialize `m_metadata` field and later use config_path in it to read data from HDFS.

src/python/deepgnn/graph_engine/snark/client.py
```diff
@@ -147,50 +147,70 @@ class MemoryGraph:
...
+            meta_path: location of meta.txt file with global graph information. If given hdfs:// or adl:// path
+                use config_path and stream parameters, for additional configuration.
+            partitions (List[Tuple[str, int]]): location of graph binary files and parition ids to load.
             storage_type (PartitionStorageType, default=memory): What type of feature / index storage to use in GE.
             config_path (str, optional): Path to folder with configuration files.
             stream (bool, default=False): If remote path is given: by default, download files first then load,
                 if stream = True and libhdfs present, stream data directly to memory -- see docs/advanced/hdfs.md for setup and usage.
         """
+        if partitions is None:
+            partitions = [(meta_path, 0)]
```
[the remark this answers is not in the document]
mortid0: Thanks! Good point!

src/python/deepgnn/graph_engine/snark/server.py
```diff
@@ -43,38 +43,55 @@ def __init__(
...
             stream (bool, default=False): If remote path is given: by default, download files first then load,
                 if stream = True and libhdfs present, stream data directly to memory.
         """
+        if partitions is None:
+            partitions = [(meta_path, 0)]
+        partitions_with_path: List[Tuple[str, int]] = []
+        for p in partitions:
+            if isinstance(p, int):
+                partitions_with_path.append((meta_path, p))
+            elif len(p) == 2:
```
coledie, line 64: Since we now support the old MemoryGraph(data_dir, [0]) can we revert all the tests and docs to use the original format.

docs/graph_engine/from_networkx.rst
```diff
@@ -82,7 +82,7 @@ Create a client to use from the temp folder:
 
 .. code-block:: python
 
-	>>> cl = client.MemoryGraph(working_dir.name, [0])
+	>>> cl = client.MemoryGraph(working_dir.name)
```
coledie, line 85: I think we should leave in the partitions [0] for this example so new users are aware of this parameter.

cl = client.MemoryGraph(working_dir.name, [0])

src/python/deepgnn/graph_engine/snark/local.py
```diff
@@ -35,11 +35,14 @@ def __init__(
         if len(partitions) == 0:
             self.logger.info("Partitions not set, defaulting to [0]")
             partitions = [0]
+        partitions_with_paths = [
```
coledie, line 38: Why does client and server not use the same version of partitions as memorygraph? Why is it List[int] where MemoryGraph is [(str, int), ...]?

coledie (changes requested):

src/python/deepgnn/graph_engine/snark/local.py
mortid0: This is a wrapper used by models. Not actual snark.Client or snark.Server.

docs/graph_engine/from_networkx.rst
mortid0: Fewer parameters is better for cognitive load.

src/python/deepgnn/graph_engine/snark/server.py
mortid0: Why? I'd rather be more explicit.

docs/graph_engine/from_networkx.rst
coledie: Yes but if the user increases the converter num_workers > 1, they will not know how to load it.

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
```diff
@@ -698,25 +768,25 @@ def test_feature_extraction_after_reset(multi_partition_graph_data):
         )
 
 
-@pytest.mark.parametrize("multi_partition_graph_data", param, indirect=True)
+@pytest.mark.parametrize("multi_partition_graph_data", ["original"], indirect=True)
```
coledie, line 771: Why is param switched to just ["original"] shouldn't this test be run on all input formats?

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
```diff
@@ -918,16 +1021,22 @@ def test_remote_client_node_extra_features(multi_partition_graph_data, storage_t
     "storage_type",
     [client.PartitionStorageType.memory, client.PartitionStorageType.disk],
 )
-@pytest.mark.parametrize("multi_partition_graph_data", param, indirect=True)
+@pytest.mark.parametrize("multi_partition_graph_data", ["original"], indirect=True)
```
coledie, line 1024: Same here

coledie (approved):

docs/graph_engine/from_networkx.rst
mortid0: They can read the docstring to class explaining parameters and for advanced usage we still have examples in `spark_converter` an `custom_decoder` docs. This is a first doc with examples in our index and we add more complex things later.

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
mortid0: It doesn't look like it was tested at all. I noticed there were some failures reported, but overall `bazel test` command printed green. It turns out python fixtures were failing and, in this case, `pytest` will not report a test failure, but rather test to be in error state. The culprit was a crash in starting a GE to test:
image
And from what I can tell these tests were [failing since they were introduced](https://github.com/microsoft/DeepGNN/actions/runs/3340633799/jobs/5531092167):
image

Hence, I made them to run on parameters they are passing, because it doesn't look like they should run with other premutation.

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
mortid0: answered above

src/python/deepgnn/graph_engine/snark/tests/e2e_test.py
coledie: I see, should this be added as a task to fix or are they inconsequential?

mortid0: Graph binaries should be the same in e2e tests and most of other tests in this module already have a filter for `["original"]` parameters, so it seems like these tests should've had this filter anyway.

mortid0: Tests are fixed now in a separate PR and everything is passing here.

PR #107, opened 2022-10-26
[2/2] Ray Dataset

coledie:

Previous Behavior
----------------
* We use custom DeepGNN dataset and sampler objects.

New Behavior
----------------
* Allow user to define ray data pipelines for usage in GE, https://docs.ray.io/en/latest/data/pipelining-compute.html#pipelining-datasets
* Add __reduce__ method to GE that will reinitialize graph engnine when it is de-serialized witihin ray data storage.
* Add delayed start parameter to MemoryGraph to delay GE load until after initialization.


src/python/deepgnn/graph_engine/snark/distributed.py
```diff
@@ -1,18 +1,37 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-"""Snark districuted client implementation."""
-from typing import List, Dict
+"""
```
mortid0, line 4: why docstring and classes are copy pasted from [synchronized.py](https://github.com/microsoft/DeepGNN/blob/main/src/python/deepgnn/graph_engine/backends/snark/synchronized.py)?

mortid0 (commented):
Is this PR a work in progress?

src/python/deepgnn/graph_engine/snark/distributed.py
coledie: Fixed

src/cc/lib/py_graph.cc
```diff
@@ -910,6 +916,12 @@ int32_t RandomWalk(PyGraph *py_graph, int64_t seed, float p, float q, NodeID def
     return 0;
 }
 
+int32_t DeleteSampler(PySampler *py_sampler)
```
mortid0, line 919: Delete methods are not needed, we already have Reset.

src/python/deepgnn/graph_engine/snark/distributed.py
```diff
@@ -16,16 +16,18 @@
...
 
-    def __init__(self, servers: List[str], ssl_cert: str = None):
+    def __init__(
+        self, servers: List[str], ssl_cert: str = None, delayed_start: bool = False
+    ):
         """Init snark client to wrapper around ctypes API of distributed graph."""
         self.logger = get_logger()
-        self.logger.info(f"servers: {servers}. SSL: {ssl_cert}")
-        self.graph = client.DistributedGraph(servers, ssl_cert)
+        # self.logger.info(f"servers: {servers}. SSL: {ssl_cert}")
```
mortid0, line 24: why logging is removed?

CHANGELOG.md
```diff
@@ -6,11 +6,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add delayed_start parameter to graph engine to enable delaying loading graph lib until after deserialization, for use in Ray data.
+
 ### Changed
 - Replace custom trainers with Ray Train, see [docs/torch/node_class.rst](https://github.com/microsoft/DeepGNN/tree/main/docs/torch/node_class.rst).
 
+- Replace DeepGNN Samplers, TorchDeepGNNDataset and Torch Dataloader usage with Ray data usage, see [docs/graph_engine/dataset.rst](https://github.com/microsoft/DeepGNN/tree/main/docs/graph_engine/dataset.rst).
```
mortid0, line 15: I don't see any of these changes in this PR.

mortid0 (changes requested):

src/python/deepgnn/graph_engine/snark/distributed.py
coledie: I added silent mode, it is because graph client gets re-initialized so many times the logs are too much.

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,185 @@
...
+
+.. code-block:: python
+
+    >>> dataset = ray.data.range(2708).repartition(2708 // 512)
+    >>> dataset
+    Dataset(num_blocks=5, num_rows=2708, schema=<class 'int'>)
+
+We convert this dataset to a data pipeline by splitting it into windows.
+
+"Dataset pipelines allow Dataset transformations to be executed incrementally
```
mortid0, line 49: Please, avoid copy pasting external documentation and describe what and *why* we are doing in your own words.

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,185 @@
+***********************************
```
mortid0, line 1: Please, add these documents to corresponding indexes so rdt can pick them up.

mortid0 (changes requested):

CHANGELOG.md
```diff
@@ -9,8 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Add usage example for Ray Train, see [docs/torch/ray_usage.rst](https://github.com/microsoft/DeepGNN/tree/main/docs/torch/ray_usage.rst).
 
+- Add documentation for Ray Data usage, see [tutorial](https://github.com/microsoft/DeepGNN/tree/main/docs/graph_engine/dataset.rst) and [example](https://github.com/microsoft/DeepGNN/tree/main/docs/graph_engine/ray_usage_advanced.rst)
+
 - Breaking. Rename get_feature_type -> get_python_type.
 
+### Fixed
+- ~1Mb leak when reset is not called before deleting python graph engine object.
```
mortid0, line 17: nit: `Implement `__del__` method to release C++ client and server. Important for ray actors, because they create numerous clients during training.`

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,201 @@
...
+Ray Dataset and Data Pipeline Usage
+***********************************
+
+In this guide we show how to create and use `Ray Datasets <https://docs.ray.io/en/latest/data/dataset.html>`
```
mortid0, line 5: Please, fix link rendering:
image

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,201 @@
...
+.. code-block:: python
+
+    >>> dataset = ray.data.range(2708).repartition(2708 // 512)
+    >>> dataset
+    Dataset(num_blocks=5, num_rows=2708, schema=<class 'int'>)
+
+We convert this dataset to a data pipeline by splitting it into windows. Each window is effectively a separate
+dataset object and the dataset pipeline consists of multiple windows which each include mulitple blocks / batches.
+When we run a function on a dataset pipeline it is not executed immediately, instead it is staged and only run
+per window when required by iter_batches.
```
mortid0, line 50: nit: highlight `iter_batches`.

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,201 @@
...
+
+.. code-block:: python
+
+    >>> def transform_batch(idx: list) -> dict:
+    ...     return {"features": g.node_features(idx, np.array([[1, 50]]), feature_type=np.float32), "labels": np.ones((len(idx)))}
+    >>> pipe = pipe.map_batches(transform_batch)
+    >>> pipe
+    DatasetPipeline(num_windows=6, num_stages=3)
+
+Finally we iterate over the dataset n_epochs times.
```
mortid0, line 93: highlight `n_epochs`

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,201 @@
...
+
+    >>> pipe = pipe.random_shuffle_each_window(seed=0)
+    >>> pipe
+    DatasetPipeline(num_windows=6, num_stages=2)
+
+Use `map_batches <https://docs.ray.io/en/latest/data/api/dataset.html#ray.data.Dataset.map_batches>`
+to map node ids from the sampler to a dictionary of node features and labels for the model forward function.
+Since this is run on the dataset pipeline, the node ids will not be mapped all at once, only when needed during iteration.
+
+For each query output vector, each first dimension needs to be equal to the batch size == len(idx).
```
mortid0, line 83: nit: just `For each query output vector, each first dimension needs to be equal to the batch size.`

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,201 @@
...
+
+.. code-block:: python
+
+    >>> def transform_batch(idx: list) -> dict:
+    ...     return {"features": g.node_features(idx, np.array([[1, 50]]), feature_type=np.float32), "labels": np.ones((len(idx)))}
```
mortid0, line 88: why does every label equal to `1`?

docs/torch/ray_usage_advanced.rst
```diff
@@ -0,0 +1,247 @@
...
+
+.. code-block:: python
+
+    >>> import tempfile
+	>>> from deepgnn.graph_engine.data.citation import Cora
+    >>> data_dir = tempfile.TemporaryDirectory()
+	>>> Cora(data_dir.name)
+	<deepgnn.graph_engine.data.citation.Cora object at 0x...>
+
+GAT Model
```
mortid0, line 18: Can you add some comments, what are we going to implement and how? It is just a wall of code right now, without any explanation what is going on:
image

docs/torch/ray_usage_advanced.rst
```diff
@@ -0,0 +1,247 @@
...
+
+Train
+=====
+
+Here we define our training function.
+In the setup part we do two notable things things,
+
+* Wrap the model and optimizer with train.torch.prepare_model/optimizer for Ray multi worker usage.
+
+* Initialize the ray dataset, see more details in `docs/graph_engine/dataset.rst`.
```
mortid0, line 149: Please, fix link rendering: 
image

mortid0: Same below.

docs/torch/ray_usage_advanced.rst
```diff
@@ -0,0 +1,247 @@
...
+    ...         "batch_size": 2708,
+    ...         "data_dir": data_dir.name,
+    ...         "sample_filename": "test.nodes",
+    ...         "n_epochs": 1,
+    ...         "model_dir": f"{model_dir.name}/model.pt",
+    ...     },
+    ...     run_config=RunConfig(verbose=0),
+    ...     scaling_config=ScalingConfig(num_workers=1, use_gpu=False, _max_cpu_fraction_per_node = 0.8),
+    ... )
+    >>> result = trainer.fit()
```
mortid0, line 244: what is the expected result, how do we know training produced a meaningful model?

docs/torch/ray_usage_advanced.rst
coledie: I'll add comments to reference main node class, in the future this will be the only of the three(node_class, ray_usage, ray_usage_advanced) so I will pull those docs from node_class later.

docs/torch/ray_usage_advanced.rst
coledie: This will be the same as in the #1, Right now I am printing metric and loss.

src/python/deepgnn/graph_engine/snark/distributed.py
```diff
@@ -84,3 +92,12 @@ def reset(self):
         """Reset server."""
         if self.server is not None:
             self.server.reset()
+
+    def __reduce__(self):
+        """On serialize reload as Client."""
```
mortid0, line 97: why do we reload server as a client? What happens to an old server?

src/python/deepgnn/graph_engine/snark/distributed.py
```diff
@@ -84,3 +92,12 @@ def reset(self):
...
+
+        def deserialize(*args):
+            get_logger().setLevel(logging.ERROR)
+            return Client(*args)
+
+        return deserialize, ([self._hostname], self._ssl_cert)
```
mortid0, line 103: How do we handle a multi-server setup?

src/python/deepgnn/graph_engine/snark/distributed.py
coledie: Server stays in main thread, when ge is serialized into plasma storage and pulled out in the worker it reloads as client.

src/python/deepgnn/graph_engine/snark/distributed.py
coledie: This will have to be done manually e.g. in the transform function just manually create a client.

mortid0: Discuss above.

src/python/deepgnn/graph_engine/snark/distributed.py
mortid0: Does the worker use this client then? If no, then why do we create it. If yes, then a multi-server setup will fail, because client will not be connected to all of them.

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,201 @@
...
+
+    >>> def transform_batch(idx: list) -> dict:
+    ...     return {"features": g.edge_features(idx, np.array([[0, 2]]), feature_type=np.float32), "labels": g.edge_features(idx, np.array([[1, 1]]), feature_type=np.float32)}
+    >>> pipe = pipe.map_batches(transform_batch)
+    >>> pipe
+    DatasetPipeline(num_windows=None, num_stages=2)
+
+    >>> batch = next(pipe.iter_torch_batches(batch_size=2))
+    >>> batch
+    {'features': tensor([[0., 0.],
```
mortid0, line 199: 1. These zeros look like they are default stubs. Can we sample something that is different?
2. Do we need to set seed to avoid random failures in a doctest?

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,201 @@
...
+
+Finally we iterate over the dataset `n_epochs` times.
+
+.. code-block:: python
+
+    >>> epoch_iter = pipe.iter_epochs()
+    >>> epoch_pipe = next(epoch_iter)
+    >>> batch = next(epoch_pipe.iter_torch_batches(batch_size=2))
+    >>> batch
+    {'features': tensor([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
```
mortid0, line 101: nit: just check first few zeros and then wildcard the rest to keep docs concise.

docs/graph_engine/dataset.rst
```diff
@@ -0,0 +1,201 @@
...
+.. code-block:: python
+
+    >>> n_epochs = 2
+    >>> pipe = pipe.repeat(n_epochs)
+    >>> pipe
+    DatasetPipeline(num_windows=6, num_stages=1)
+
+We add shuffling at this part of the dataset so that we only shuffle node ids, not the whole query return. It is important to add this after repeat so it does not get cached.
+
+    >>> pipe = pipe.random_shuffle_each_window(seed=0)
```
mortid0, line 75: do we need to set seed?

src/python/deepgnn/graph_engine/snark/distributed.py
coledie: Yes, it makes the code simpler for the basic user. e.g.

```python
g = Server(address, host)
def transform(batch):
    g # Here this is actually a client
    return model.query(g, batch)
pipeline.map_batches(transform)
```
Otherwise for multi serveres they'll do
```python
s = [Server(address, host), Server(address, host)]
def transform(batch):
    g = DistributedClient([address1, address2])
    return model.query(g, batch)
pipeline.map_batches(transform)
```

mortid0: The transform function should be independent from data setup though.

docs/graph_engine/dataset.rst
coledie: This one is for the edge sampler, I'd have to make a custom dataset to avoid this.

docs/graph_engine/dataset.rst
coledie: Yes

src/python/deepgnn/graph_engine/snark/distributed.py
coledie: Multiple servers can do this instead to make it independent of setup. I'll add test case for this in dataset.rst
```python
s = [Server(address, host), Server(address, host)]
g = DistributedClient([address1, address2])
def transform(batch):
    return model.query(g, batch)
pipeline.map_batches(transform)
```

mortid0: well, then it means we don't need to create a client on server deserialize at all.

src/python/deepgnn/graph_engine/snark/distributed.py
```diff
@@ -45,6 +63,9 @@ def __init__(
         stream: bool = False,
     ):
         """Init snark server."""
+        self._hostname = hostname
```
mortid0, line 66: why do we still need this?

requirements.txt
```diff
@@ -167,7 +167,7 @@ ptyprocess==0.7.0
 pure_eval==0.2.2
 py==1.11.0
 py4j==0.10.9.7
-pyarrow==9.0.0
```
mortid0, line 170: why are we downgrading pyarrow?

coledie: Ray requires this version of pyarrow.

mortid0 (approved):

PR #112, opened 2022-11-03
Update converter docs.

coledie:

New Behavior
----------------
Update converter docs per user request.

### DIFF
--- a/src/python/deepgnn/graph_engine/snark/convert.py
+++ b/src/python/deepgnn/graph_engine/snark/convert.py
@@ -50,8 +50,8 @@ class MultiWorkersConverter:
             output_dir: the output directory to put the generated graph binary files.
             decoder (Decoder): Decoder object which is used to parse the raw graph data file.
             partition_count: how many partitions will be generated.
-            worker_index: the work index when running in multi worker mode.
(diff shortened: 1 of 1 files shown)


src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -50,8 +50,8 @@ def __init__(
             output_dir: the output directory to put the generated graph binary files.
             decoder (Decoder): Decoder object which is used to parse the raw graph data file.
             partition_count: how many partitions will be generated.
-            worker_index: the work index when running in multi worker mode.
-            worker_count: how many workers will be started to convert the data, use 0 with 1 partition for debug mode.
+            worker_index: If using multiple MultiWorkersConverters, worker_index is the id of this converter.
+            worker_count: If using multiple MultiWorkersConverters, worker_count is the number of converters.
```
mortid0, line 54: Nit: keep message about debug mode?

mortid0 (approved):

src/python/deepgnn/graph_engine/snark/convert.py
coledie: That was actually wrong, we use the debug parameter for debug only.

PR #137, opened 2023-01-05
Add TF version of Ray usage.

coledie:

New Behavior
----------------
Add documentation for example on how to use Ray with Tensorflow.
-> This is a repeat of a previous pull request, this is meant to merge in with the torch ray_usage docs before the migration.


docs/tf/ray_usage.rst
```diff
@@ -0,0 +1,260 @@
+**************************************************
+Ray Usage Example for Node Classification with GAT
+**************************************************
+
+.. code-block:: python
```
mortid0, line 5: Please add some comments to the code. It is not self-explanatory for someone who starts directly with deepgnn and TF.

mortid0 (changes requested):

docs/tf/ray_usage.rst
coledie: This is a copy of the node class example, only with small changes to make it use Ray.

mortid0: A person who works with TF doesn't know about a pytorch copy and that they should start with it. Also why should they?

coledie: That should be for a different PR then, this is meant to be a new version of the tf_node_class that uses Ray. The original version has no documentation either,

mortid0 (approved):

PR #138, opened 2023-01-06
[1/2] Add example of migrate data.

coledie:

New Behavior
----------------
Add example for what pytorch examples will look like after migrate script.


examples/pytorch/gat/main.py
```diff
@@ -10,28 +10,11 @@
 from deepgnn.pytorch.common.dataset import TorchDeepGNNDataset
```
coledie, line 10: Deltas on this file don't matter, it is the same as it was before. This PR is for example of migrate script not final examples.

coledie:
First PR does not include Ray Data, next one will

PR #139, opened 2023-01-06
Save 10 mins in lint

coledie:

New Behavior
----------------
Remove tensorflow and ray from macos pip installs for lint because they require manual building of grpcio which is very slow.

### DIFF
--- a/.github/actions/lint/action.yml
+++ b/.github/actions/lint/action.yml
@@ -15,6 +15,11 @@ runs:
     - run: pip install wheel pre-commit==2.17.0 mypy==0.971 numpy==1.22.4 torch==1.11.0 tensorflow==2.8.3 ray==2.2.0
       shell: bash
       name: install dependencies
+      if: runner.os != 'macOS'
+    - run: pip install wheel pre-commit==2.17.0 mypy==0.971 numpy==1.22.4 torch==1.11.0
+      shell: bash
+      name: install dependencies
+      if: runner.os == 'macOS'
     - run: pre-commit install
       shell: bash
       name: initialize pre-commit


.github/actions/lint/action.yml
```diff
@@ -15,6 +15,11 @@ runs:
     - run: pip install wheel pre-commit==2.17.0 mypy==0.971 numpy==1.22.4 torch==1.11.0 tensorflow==2.8.3 ray==2.2.0
       shell: bash
       name: install dependencies
+      if: runner.os != 'macOS'
+    - run: pip install wheel pre-commit==2.17.0 mypy==0.971 numpy==1.22.4 torch==1.11.0
+      shell: bash
+      name: install dependencies
+      if: runner.os == 'macOS'
```
mortid0, line 22: We already ignore all extra packages in .mypy file. Please, simply remove `numpy==1.22.4 torch==1.11.0 tensorflow==2.8.3 ray==2.2.0`.

mortid0 (approved):

.github/actions/lint/action.yml
coledie: mypy only ignores missing imports, these are still needed actually there is a PR or two from a while ago.

mortid0: It ignores everything:
diff: https://github.com/mortid0/DeepGNN/commit/921f7226f365e30e593342ee8bcbecc58bff82b2
run: https://github.com/mortid0/DeepGNN/actions/runs/3858029307

coledie: Thats because I fixed all the bugs from it in a previous pull request already: https://github.com/microsoft/DeepGNN/pull/75/files

mortid0: Do we still have to install numpy, torch, TF etc? If yes, then we should start to enforce it. If no, then remove them.

coledie: I can do this but that belongs in another PR.

mortid0: Could you elaborate what do you mean by ‘this’?

coledie: REmoving the mypy rules for numpy, ...

PR #145, opened 2023-01-11
Add grpc options

mortid0:

Previous Behavior
----------------
We had only a small number of settings to configure networking.

New Behavior
----------------
Tuple of key value strings is passed to a client, so userscan control networking directly. Similar to [this](
https://github.com/grpc/grpc/blob/2caa79feeb53c982e7afad1a4f872cdbf166aefd/examples/python/retry/retry_client.py#L25).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -161,7 +161,7 @@ portalocker==2.5.1
 prettytable==3.4.1
 prometheus-client==0.14.1
 prompt-toolkit==3.0.31
-protobuf==4.21.12
+protobuf==3.20.3
 psutil==5.9.2
 ptyprocess==0.7.0
 pure_eval==0.2.2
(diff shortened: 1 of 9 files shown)


requirements.txt
```diff
@@ -161,7 +161,7 @@ portalocker==2.5.1
 prettytable==3.4.1
 prometheus-client==0.14.1
 prompt-toolkit==3.0.31
-protobuf==3.20.3
+protobuf==4.21.12
```
coledie, line 164: Is it possible to avoid this? Ray requires a version <= 3.20

coledie (approved):

requirements.txt
mortid0: I think so.

PR #148, opened 2023-01-18
Improve doctest stability.

coledie:

Previous Behavior
----------------
* Ray doctest keeps failing.

New Behavior
----------------
* TF docs use > % now instead of fixed since it is not possible to seed tf/keras properly.
* Switch to jobs == 1 to prevent ray starting multiple clusters at once, does not make a diff to runtime since only 2 cpus in runner.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -19,7 +19,6 @@ build:linux --action_env=SETUPTOOLS_USE_DISTUTILS="false"
 build:linux --action_env=HADOOP_HOME="/tmp/snark/"
 build:linux --action_env=SNARK_LIB_PATH="./src/cc/lib/libwrapper.so"
 build:linux --action_env=LD_LIBRARY_PATH="./external/jvm/jre/lib/amd64/server"
-build:linux --action_env=PYTHONHASHSEED=0
 
 build:manylinux --crosstool_top=//tools/toolchain:manylinux
 build:manylinux --cpu=k8
--- a/.bazelrc
+++ b/.bazelrc
@@ -19,6 +19,7 @@ build:linux --action_env=SETUPTOOLS_USE_DISTUTILS="false"
 build:linux --action_env=HADOOP_HOME="/tmp/snark/"
 build:linux --action_env=SNARK_LIB_PATH="./src/cc/lib/libwrapper.so"
 build:linux --action_env=LD_LIBRARY_PATH="./external/jvm/jre/lib/amd64/server"
+build:linux --action_env=PYTHONHASHSEED=0
 
 build:manylinux --crosstool_top=//tools/toolchain:manylinux
 build:manylinux --cpu=k8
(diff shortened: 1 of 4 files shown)


docs/tf/ray_usage.rst
```diff
@@ -250,11 +250,11 @@ Ray Usage Example for Node Classification with GAT
     ...     train_loop_config={
     ...         "batch_size": 2708,
     ...         "data_dir": data_dir.name,
-    ...         "n_epochs": 100,
+    ...         "n_epochs": 300,
```
mortid0, line 253: why do we need this? I thought training should be completely deterministic.

coledie: Not in tf/keras, it is not possible to properly seed it but higher epochs helps with stability.

mortid0: https://keras.io/getting_started/faq/#how-can-i-obtain-reproducible-results-using-keras-during-development

coledie: Ah even better thank you

coledie: This improves stability but still 1/3 is == .5.

.bazelrc
```diff
@@ -19,6 +19,7 @@ build:linux --action_env=SETUPTOOLS_USE_DISTUTILS="false"
 build:linux --action_env=HADOOP_HOME="/tmp/snark/"
 build:linux --action_env=SNARK_LIB_PATH="./src/cc/lib/libwrapper.so"
 build:linux --action_env=LD_LIBRARY_PATH="./external/jvm/jre/lib/amd64/server"
+build:linux --action_env=PYTHONHASHSEED=0
```
mortid0, line 22: Please, use env field only for test targets that need it:
https://bazel.build/reference/be/common-definitions#common-attributes-tests

docs/tf/ray_usage.rst
```diff
@@ -250,11 +250,11 @@ Ray Usage Example for Node Classification with GAT
...
-    ...         "n_epochs": 100,
+    ...         "n_epochs": 120,
     ...     },
     ...     run_config=RunConfig(verbose=0),
     ...     scaling_config=ScalingConfig(num_workers=1, use_gpu=False),
     ... )
     >>> result = trainer.fit()
-    >>> result.metrics["accuracy"]
-    [0.8...]
+    >>> result.metrics["accuracy"][0] > .5
```
mortid0, line 259: why can't we check an actual value with 1% accuracy at least?

coledie: The test still is not stable it varies by about 30%.

mortid0: That's a lot and it means we can't use it for documentation. Please either stabilize or remove it.

mortid0 (approved):

PR #149, opened 2023-01-18
[Temporal graph][4/4] Converter

mortid0:

New Behavior
----------------
1. Update converter to process temporal elements of the graph: nodes/edges/features with "created_at" and "deleted_at".
2. Added converter and e2e tests in python
3. Updated ppr-sampling to work for temporal graphs and docs

### DIFF
--- a/docs/graph_engine/custom_decoder.rst
+++ b/docs/graph_engine/custom_decoder.rst
@@ -85,11 +85,11 @@ index 0 and 2 but skip index 1.
     ...         # Args can be added to init and added before passing to MultiWorkersConverter
     ...         pass
     ...
-    ...     def decode(self, line: dict) -> Iterator[Tuple[int, int, int, float, list]]:
-    ...         # For a node we will yield: (node_id, -1, node_type, node_weight, [feature_0, ...])
-    ...         # For an edge: (src_id, dst_id, edge_type, edge_weight, [feature_0, ...])
+    ...     def decode(self, line: dict) -> Iterator[Tuple[int, int, int, float, int, int, list]]:
+    ...         # For a node we will yield: (node_id, -1, node_type, node_weight, None, None, [feature_0, ...])
+    ...         # For an edge: (src_id, dst_id, edge_type, edge_weight, None, None, [feature_0, ...])
     ...         # We can return multiple items per line, but the order of the file, described above, must be maintained.
-    ...         yield int(line["src"]), int(line["dst"]), int(line["type"]), float(line["weight"]), [np.array(line["float_feature"], dtype=np.float32), (np.array(line["sparse_int16_coords"], dtype=np.int64), np.array(line["sparse_int16_feature"], dtype=np.int16))]
+    ...         yield int(line["src"]), int(line["dst"]), int(line["type"]), float(line["weight"]), None, None, [np.array(line["float_feature"], dtype=np.float32), (np.array(line["sparse_int16_coords"], dtype=np.int64), np.array(line["sparse_int16_feature"], dtype=np.int16))]
 
 Finally, we convert our input file to binaries using an Avro reader, a BinaryWriter and the AvroDecoder.
 Here we only have one partition and therefore one binary writer, it is okay to have multiple binary writers
@@ -123,9 +123,10 @@ Here we manually write a meta.json file for our graph engine to load.
     ...         "partitions": {"0": {"node_weight": writer.node_weight, "edge_weight": writer.edge_weight}},
     ...         "node_count_per_type": writer.node_type_count,
     ...         "edge_count_per_type": writer.edge_type_count,
+    ...         "watermark": -1,
     ...     }
     ...     f.write(json.dumps(content))
-    280
+    297
 
 We load the generated binaries into a graph engine and demonstrate it working.
 
--- a/src/python/deepgnn/graph_engine/snark/convert.py
+++ b/src/python/deepgnn/graph_engine/snark/convert.py
@@ -40,6 +40,7 @@ class MultiWorkersConverter:
         skip_edge_sampler: bool = False,
         file_iterator: Optional[TextFileIterator] = None,
         debug: bool = False,
+        watermark: Optional[int] = None,
     ):
         """Run multi worker converter in multi process.
 
@@ -62,6 +63,7 @@ class MultiWorkersConverter:
             skip_edge_sampler(bool): skip generation of edge alias tables.
             file_iterator(TextFileIterator): Iterator to yield lines of the input text file.
             debug(bool, False): Enable debug mode to disable multiprocessing and see error messages, forces worker_count=1, paritition_count=1.
+            watermark(Optional[int]): for temporal graphs is latest known timestamp in the graph.
         """
         if decoder is None:
             decoder = JsonDecoder()  # type: ignore
@@ -77,6 +79,7 @@ class MultiWorkersConverter:
         self.thread_count = thread_count
         self.dispatcher = dispatcher
         self.file_iterator = file_iterator
+        self.watermark = watermark
 
         self.fs, _ = get_fs(graph_path)
 
@@ -103,6 +106,7 @@ class MultiWorkersConverter:
                 use_threads=use_threads,
                 skip_node_sampler=skip_node_sampler,
                 skip_edge_sampler=skip_edge_sampler,
+                watermark=watermark,
             )
 
     def convert(self):
@@ -143,9 +147,14 @@ class MultiWorkersConverter:
             ), "Debug mode does not support multiple partitions."
             if isinstance(self.decoder, type):
                 self.decoder = self.decoder()
-            writer = BinaryWriter(self.output_dir, 0)
+            writer = BinaryWriter(
+                self.output_dir,
+                0,
+                watermark=self.watermark,
+            )
             for _, data in enumerate(self.file_iterator):
                 for line in data:
+                    print(f"lise is {line}")
                     writer.add(self.decoder.decode(line))
             writer.close()
             gettr = lambda key: getattr(writer, key)  # noqa: E731
@@ -184,6 +193,7 @@ class MultiWorkersConverter:
 
         mjson["node_count_per_type"] = node_count_per_type
         mjson["edge_count_per_type"] = edge_count_per_type
+        mjson["watermark"] = -1 if self.watermark is None else self.watermark
 
         with fs.open(
             "{}/meta{}.json".format(
@@ -235,6 +245,12 @@ if __name__ == "__main__":
         default=False,
         help="Skip generation of edge alias tables for edge sampling",
     )
+    parser.add_argument(
+        "--watermark",
+        type=int,
+        default=-1,
+        help="Watermark(latest timestamp) to use for temporal graphs. -1 for other graphs",
+    )
     args = parser.parse_args()
 
     decoder = getattr(decoders, f"{args.type.capitalize()}Decoder")()
@@ -250,5 +266,6 @@ if __name__ == "__main__":
         decoder=decoder,
         skip_node_sampler=args.skip_node_sampler,
         skip_edge_sampler=args.skip_edge_sampler,
+        watermark=args.watermark,
     )
     c.convert()
--- a/src/python/deepgnn/graph_engine/snark/converter/process.py
+++ b/src/python/deepgnn/graph_engine/snark/converter/process.py
@@ -28,17 +28,19 @@ def converter_process(
     decoder: DecoderType,
     skip_node_sampler: bool,
     skip_edge_sampler: bool,
+    watermark: typing.Optional[int] = None,
 ) -> None:
     """Process graph nodes from a queue to binary files.
 
     Args:
-        q_in (typing.Union[mp.Queue, Connection]): input nodes
-        q_out (mp.Queue): signal processing is done
-        folder (str): where to save binaries
-        suffix (int): file suffix in the name of binary files
+        q_in (typing.Union[mp.Queue, Connection]): input nodes.
+        q_out (mp.Queue): signal processing is done.
+        folder (str): where to save binaries.
+        suffix (int): file suffix in the name of binary files.
         decoder (Decoder): Decoder object which is used to parse the raw graph data file.
-        skip_node_sampler(bool): skip generation of node alias tables
-        skip_edge_sampler(bool): skip generation of edge alias tables
+        skip_node_sampler (bool): skip generation of node alias tables.
+        skip_edge_sampler (bool): skip generation of edge alias tables.
+        watermark (optional[int]): latest known timestamp for temporal graphs.
     """
     if isinstance(decoder, type):
         decoder = decoder()
@@ -48,6 +50,7 @@ def converter_process(
         suffix,
         skip_node_sampler,
         skip_edge_sampler,
+        watermark,
     )
     while True:
         if type(q_in) == Connection:
--- a/src/python/deepgnn/graph_engine/snark/converter/writers.py
+++ b/src/python/deepgnn/graph_engine/snark/converter/writers.py
@@ -31,6 +31,7 @@ class BinaryWriter:
         suffix: int,
         skip_node_sampler: bool = False,
         skip_edge_sampler: bool = False,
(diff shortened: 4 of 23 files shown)


src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         skip_edge_sampler: bool = False,
         file_iterator: Optional[TextFileIterator] = None,
         debug: bool = False,
+        is_temporal: bool = False,
```
coledie, line 42: We should auto-detect this instead to keep it simple for users.

src/python/deepgnn/graph_engine/snark/tests/temporal_test.py
```diff
@@ -0,0 +1,352 @@
...
+    data_name = triangle_graph_json(output.name)
+    d = dispatcher.QueueDispatcher(Path(output.name), 2, Counter(), JsonDecoder())
+    convert.MultiWorkersConverter(
+        graph_path=data_name,
+        partition_count=1,
+        output_dir=output.name,
+        decoder=JsonDecoder(),
+        dispatcher=d,
+        debug=True,
+    ).convert()
```
coledie, line 166: Should have is_temporal=True?

src/python/deepgnn/graph_engine/snark/tests/temporal_test.py
```diff
@@ -0,0 +1,352 @@
+# Copyright (c) Microsoft Corporation.
```
coledie, line 1: I think there should be updates to the converter test case for this update.

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -275,7 +283,7 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
 
             features.append(self._get_feature(key, length, data))
 
-        yield src, dst, typ, weight, features
+        yield src, dst, typ, weight, features, None, None
```
coledie, line 286: Edge list should support temporal graph not just json.

coledie (changes requested):

src/python/deepgnn/graph_engine/snark/convert.py
mortid0: How can we autodetect it, if we also want to support default values?

src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -39,6 +39,8 @@ def __init__(
...
+        watermark: int = -1,
```
coledie, line 43: Can we merge these two parameters? e.g. if watermark != -1 then is_temporal is true.

src/python/deepgnn/graph_engine/snark/convert.py
```diff
@@ -61,6 +63,8 @@ def __init__(
             skip_edge_sampler(bool): skip generation of edge alias tables.
             file_iterator(TextFileIterator): Iterator to yield lines of the input text file.
             debug(bool, False): Enable debug mode to disable multiprocessing and see error messages, forces worker_count=1, paritition_count=1.
+            is_temporal(bool): add timestamps to graph elements.
+            watermark(int): latest known timestamp in the graph.
```
coledie, line 67: Can we rename watermark to temporal_watermark or something more clear / memorable.

src/python/deepgnn/graph_engine/snark/converter/process.py
```diff
@@ -39,6 +41,7 @@ def converter_process(
         decoder (Decoder): Decoder object which is used to parse the raw graph data file.
         skip_node_sampler(bool): skip generation of node alias tables
         skip_edge_sampler(bool): skip generation of edge alias tables
+        is_temporal(bool): process timestamps for graph elements
```
coledie, line 44: missing watermark docstring.

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -39,24 +41,22 @@ def __init__(
             suffix (int): file suffix in the name of binary files
             skip_node_sampler(bool): skip generation of node alias tables
             skip_edge_sampler(bool): skip generation of edge alias tables
+            skip_timestamps(bool): skip generation of dynamic graph data
```
coledie, line 44: missing is_temporal and watermark docstring. Also is skip_timestamps even a parmaeter?

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -39,24 +41,22 @@ def __init__(
...
         """
         self.folder = folder
         self.suffix = suffix
         self.skip_node_sampler = skip_node_sampler
         self.skip_edge_sampler = skip_edge_sampler
+        self.is_temporal = is_temporal
 
-        self.node_writer = NodeWriter(self.folder, self.suffix)
-        self.edge_writer = EdgeWriter(self.folder, self.suffix)
+        self.node_writer = NodeWriter(folder, suffix, is_temporal)
```
coledie, line 52: Why does nodewriter not use watermark?

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -70,17 +70,33 @@ def __init__(
...
         """
         Write binary for a node and all of its edges.
 
         args:
-            data: Iterable[(int, int, int, float, list)] Data for a node first, then all of
+            data: Iterable[(int, int, int, float, list, typing.Optional[typing.Tuple[int, int]])] Data for a node first, then all of
                 its edges in order of type then dst. All of the node's outgoing edges do not need to go to the same writer.
                 Each entry for a node/edge is,
-                (node_id/src, -1/dst, type, weight, [ndarray for each feature vector or None in order of feature index]).
+                (node_id/src, -1/dst, type, weight, [ndarray for each feature vector or None in order of feature index], timestamps of creation and deletion of data).
```
coledie, line 94: Does converter have update for how feature can be created and deleted? if so documenet

src/python/deepgnn/graph_engine/snark/converter/writers.py
```diff
@@ -70,17 +70,33 @@ def __init__(
...
         """
-        for src, dst, typ, weight, features in data:
+        # Default values for timestamps: node existed since the very beginning and towards the very end.
+        node_created_at = 0
+        node_removed_at = -1
+        for src, dst, typ, weight, features, created_at, removed_at in data:
```
coledie, line 99: Won't this error if created at and removed_at are not specified? Either they should not be marked optional or this is a bug, I think this is a bug.

coledie, line 99: Also docs say it should be "src, dst, typ, weight, features, (created_at, removed_at)"

src/python/deepgnn/graph_engine/snark/decoders.py
```diff
@@ -436,7 +513,7 @@ def decode(self, line: str) -> Iterator[Tuple[int, int, int, float, list]]:
         node_weight = float(columns[2]) if columns[2] else 0.0
         yield node_id, -1, node_type, node_weight, self._parse_feature_string(
             columns[3]
-        )
+        ), None, None
```
coledie, line 516: I think we should remove the ", None, None" since these are supposed to be optional.

coledie (changes requested):

src/python/deepgnn/graph_engine/snark/convert.py
mortid0: Watermark is a standard term https://en.wikipedia.org/wiki/Watermark_(data_synchronization)

src/python/deepgnn/graph_engine/snark/decoders.py
mortid0: Added support to edge list

src/python/deepgnn/graph_engine/snark/converter/writers.py
mortid0: added docstring

src/python/deepgnn/graph_engine/snark/converter/writers.py
mortid0: created and deleted are just tags for values. It is documented in the temporal.rst file in the very first PR of this stack.

src/python/deepgnn/graph_engine/snark/decoders.py
mortid0: Optional type means the value can be `None`, not that it can be ommited from result list.

src/python/deepgnn/graph_engine/snark/converter/writers.py
mortid0: added watermark

src/python/deepgnn/graph_engine/snark/converter/writers.py
mortid0: no, tuple has to have 7 values. Optional doesn't affect number of arguments, an individual argument can be None.

coledie: You can make it so it can have 5 or 7 arguments by unrolling the tuple inside of the for loop instead, this will make it nicer and less confusing for non temporal users.

mortid0: pushed created_at and removed_at in front of the feature list. Otherwise, it can't be supported in edge list converter.

src/python/deepgnn/graph_engine/snark/convert.py
mortid0: Merged

src/python/deepgnn/graph_engine/snark/tests/temporal_test.py
mortid0: Added more tests in the converter_tests.py

src/python/deepgnn/graph_engine/snark/converter/writers.py
mortid0: It is already supported for edgelist users through default arguments:
image

Adding support for variable tuples is not part of this PR, otherwise we should've done this for graphs with same weights/types or without any features.

coledie: Put it in the changelog then

docs/graph_engine/custom_decoder.rst
```diff
@@ -85,11 +85,11 @@ index 0 and 2 but skip index 1.
     ...         # Args can be added to init and added before passing to MultiWorkersConverter
     ...         pass
     ...
-    ...     def decode(self, line: dict) -> Iterator[Tuple[int, int, int, float, list]]:
+    ...     def decode(self, line: dict) -> Iterator[Tuple[int, int, int, float, int, int, list]]:
     ...         # For a node we will yield: (node_id, -1, node_type, node_weight, [feature_0, ...])
     ...         # For an edge: (src_id, dst_id, edge_type, edge_weight, [feature_0, ...])
```
coledie, line 90: Update docs here to show created_at and deleted_at

docs/graph_engine/custom_decoder.rst
```diff
@@ -85,11 +85,11 @@ index 0 and 2 but skip index 1.
...
     ...         # We can return multiple items per line, but the order of the file, described above, must be maintained.
-    ...         yield int(line["src"]), int(line["dst"]), int(line["type"]), float(line["weight"]), [np.array(line["float_feature"], dtype=np.float32), (np.array(line["sparse_int16_coords"], dtype=np.int64), np.array(line["sparse_int16_feature"], dtype=np.int16))]
+    ...         yield int(line["src"]), int(line["dst"]), int(line["type"]), float(line["weight"]), None, None, [np.array(line["float_feature"], dtype=np.float32), (np.array(line["sparse_int16_coords"], dtype=np.int64), np.array(line["sparse_int16_feature"], dtype=np.int16))]
```
coledie, line 92: edge list needs to support temporal features as well.

coledie (changes requested):

docs/graph_engine/custom_decoder.rst
mortid0: it supports it.

coledie: What? Just make if user puts in a 3 tuple (ndarray, int, int) then that is a temporal feature

mortid0: There are no ndarrays in text format. You are mixing decoder interface and parsing.

coledie: I see,
features[dense]: dtype,length,v1,v2,...,dtype2,length2,v1,v2,...
features[sparse]: dtype,values_length/coords_dim,c1,c2,...,v1,v2,...
features[temporal_dense]: dtype,length/created_at/deleted_at,v1,v2,...
features[temporal_sparse]: dtype,values_length/coords_dim/created_at/deleted_at,c1,c2,...,v1,v2,...

mortid0: what is this?

coledie: This is a way to update the text format to include temporal features without conflicts

coledie: features[temporal_dense]: dtype,length/created_at/deleted_at,v1,v2,...
E.g. 
float32,3/0/100,0,1,2
= data=np.array([0, 1, 2], dtype=float32), created_at=0, deleted_at=100

mortid0: Feature values that are created and removed multiple times. Example from doc:
```json
"sparse_int32_feature": {
	"0": [
		{"coordinates": [[5, 13], [7, 25]], "values": [-1, 1024], "removed_at": 21},
		{"coordinates": [[4, 2], [1, 3]], "values": [2, 4], "created_at": 21}
	],
	"1": {"coordinates": [13, 42], "values": [1, 1]}
}
```

coledie: Is there something blocking us from just having multiple entries per feature id in the edge list format?

mortid0: [Features will be given indexes starting at 0 and indexes can be skipped with 0 length vectors.](https://github.com/microsoft/DeepGNN/blob/0abd02f3c11ea8e051b6d224f8a33954b7d3d90c/src/python/deepgnn/graph_engine/snark/decoders.py#L64)

coledie: features[temporal_dense]: dtype,length/created_at/deleted_at,v1,v2,...length2/created_at2/deleted_at2,v1,v2,...
Instead of having next dtype just have another tuple and it will recognize as a second temporal value.

mortid0: There are no tuples in text. Please write explicit contents of a graph with the node feature from above and state why does it have to be added in this PR.

mortid0: > Parsing needs to be added to this PR since EdgeList is the fastest format for conversion that we support and it needs to be fully featured.

This is a circular logic. `A` has to be added, because we have to have `A`. I am also pretty sure [the avro decoder in docs](https://github.com/microsoft/DeepGNN/blob/main/docs/graph_engine/custom_decoder.rst) is faster than text parsing with edge lists, because the former is binary.

> int32,2/2/0/21,5,13,7,25,-1,1024,2/2/21/100,4,2,1,3,2,4,int32,2/1,13,42,1,1

The proposed line doesn't match json example, there are no `created_at` value in the first feature(with values -1, 1024) and `removed_at` in the second(with values 2, 4). Supporting default values in positional encoding is a hard problem. Because the next item,

> int32,2/1,13,42,1,1

is indistinguishable from a dense feature, with `created_at 2`, `removed_at` time 1 and 4 feature values [13,42,1,1].

tl;dr In the current version edge list format is ambiguous for temporal features, hence we can't use it. If anyone requests it, we can design it properly and add in the future, but I don't see a reason to add it right now.

coledie (approved):

mortid0:
Merging with https://github.com/microsoft/DeepGNN/pull/176

PR #152, opened 2023-01-20
[3/3] Synchronize Server-Client using Ray

coledie:

Previous Behavior
----------------

New Behavior
----------------
* Add synchronize server using ray named actors instead of lock files.
* Server is always synchronized if ray is enabled.
* Client must connect to lock "server_state" if it wants to synchronize.


src/python/deepgnn/graph_engine/snark/distributed.py
```diff
@@ -104,7 +108,20 @@ def __init__(
             stream,  # type: ignore
         )
 
+        self.state = set_server_state(hostname, index, namespace)
+
     def reset(self):
         """Reset server."""
         if self.server is not None:
+            if self.state is not None and not ray.get(self.state.safe_to_terminate.remote()):
```
coledie, line 116: TODO need to halt until all servers done.

src/python/deepgnn/graph_engine/snark/synchronized2.py
```diff
@@ -0,0 +1,95 @@
+# Copyright (c) Microsoft Corporation.
```
coledie, line 1: NOTE This is old code for backup

mortid0: Please, remove old code, so it is visible in PR.

src/python/deepgnn/graph_engine/snark/distributed.py
```diff
@@ -6,12 +6,13 @@
 from typing import Optional, List, Dict, Union, Tuple
 import logging
 import tempfile
-
+import ray
```
mortid0, line 9: We definitely don't want the graph engine package to depend on ray.

mortid0 (changes requested):
I suggest to write an example with a distributed workflow: 
1. fetch a big graph
2. convert it in parallel to a binary format
3. start 2 GE servers
4.  create e.g. 3 workers
5. trains a small sage/gat/... model
6. when 5 is done, use that model to generate embeddings

src/python/deepgnn/graph_engine/snark/synchronized.py
```diff
@@ -0,0 +1,49 @@
...
+    return
+
+
+servers = start_servers.bind(2)
+clients = start_clients.bind(servers, 3)
+output = train.bind(servers, clients)
+
+from ray import workflow
+
+print(workflow.run(output))
```
coledie, line 49: Here is the workflow example you are looking for - start 2 servers, connect with 3 clients then train.

mortid0: 1. I don't think it is in an example folder..
2.  There is no training..
3. > We definitely don't want the graph engine package to depend on ray.

coledie: Can you review it to make sure the design is ok before I add training and a whole example?

mortid0: Could you verify that the proposed design will work on tests from:
https://github.com/microsoft/DeepGNN/blob/main/src/python/deepgnn/graph_engine/backends/snark/test_snark_client.py
first? Then try to make sure, that *graph engine package will not depend on ray* for every use case and we have an example, how to use it in a distributed setting.

coledie: It looks like the design will work on that old unit test, the new unit tests are in test_synchronized. As far as depending on Ray, the synchronized code depends on Ray since it uses Ray workflow but nothing else.

mortid0: Sounds good then. Try to autodetect host names for different settings listed in that test and graduate this PR from wip.

mortid0: I don't see this:
> Try to autodetect host names for different settings listed in that test

mortid0 (changes requested):
Please, include this new feature in an example to show how to use it.

src/python/deepgnn/graph_engine/snark/synchronized.py
coledie: What does that mean? I don't see anything in the old code about this

mortid0: We automatically detect where servers are running based on runtime configuration. Typically, 1 GE server per host where workers run with GE present on a worker with `local_rank = 0`. This unit test lists possible scenarios:
https://github.com/microsoft/DeepGNN/blob/main/src/python/deepgnn/graph_engine/backends/snark/test_snark_client.py

src/python/deepgnn/graph_engine/snark/synchronized.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Synchronized server - client setup."""
+from deepgnn.graph_engine.snark.distributed import Server, Client as DistributedClient
+import ray
```
mortid0, line 5: This creates an explicit dependency on ray for deepgnn-ge package. We want to avoid it and keep dependencies minimal. Please, make sure it is an [extra instead](https://peps.python.org/pep-0508/#extras).

mortid0 (changes requested):
Neither dependency problem, nor autodetection of where to run GE servers based on runtime configuration is addressed.

src/python/deepgnn/graph_engine/snark/synchronized.py
coledie: As long as the user does not import synchronized.py, they are not required to have Ray - e.g. all the unit tests work even though ray is not specified as a dependency.

src/python/deepgnn/graph_engine/snark/synchronized.py
coledie: So if a user sets "OMPI_COMM_WORLD_LOCAL_RANK": "0",
as an environment variable we automatically set the hostname to the horovod default IP?

mortid0: yes

src/python/deepgnn/graph_engine/snark/synchronized.py
coledie: We no longer have backends, we just use Ray data directly on servers.

mortid0: Sure, but I need to see how it is done. So far, backend code is still [present in this branch](https://github.com/microsoft/DeepGNN/tree/cdieckhaus/synchronize/src/python/deepgnn/graph_engine/backends).

docs/advanced/synchronize.rst
```diff
@@ -0,0 +1,57 @@
...
+
+.. code-block:: python
+
+    >>> servers = start_servers.bind("localhost:9999", data_dir.name, 2, 2)
+
+Start 3 clients
+
+.. code-block:: python
+
+    >>> clients = start_clients.bind(servers, 3)
```
coledie, line 50: make n_client == num train workers

coledie, line 50: Make servers start in separate processess

src/python/deepgnn/graph_engine/snark/synchronized.py
coledie: This is only for TF support, it is fully removed from pytorch at this point. What do you want me to update the code to do?

mortid0: In the current form it is more confusing to have multiple synchronized modules. So I would rather not create this one at all, because it is not used in any example and documentation is contrived.

coledie: Close this PR?

mortid0: Yes. Synchronized module can be updated after tf is migrated.

PR #154, opened 2023-01-26
[2/3] Examples use Ray argument passing instead of command lines.

coledie:
- [X] PR is labeled using the label menu on the right side.

Previous Behavior
----------------
Examples use command line arguments.

New Behavior
----------------
Examples now use ray argument passing.


examples/pytorch/gat/main.py
```diff
@@ -133,17 +118,46 @@ def transform_batch(idx: list) -> dict:
...
+            "sample_file": "/tmp/cora/train.nodes",
+            "partitions": 1,
+            "num_epochs": 180,
+            "batch_size": 140,
+            "feature_idx": 0,
+            "feature_dim": 1433,
+            "label_idx": 1,
+            "label_dim": 1,
+            "num_classes": 7,
+            "mode": "train",
```
mortid0, line 138: Mode is not needed

examples/pytorch/gat/main.py
```diff
@@ -133,17 +118,46 @@ def transform_batch(idx: list) -> dict:
...
     ray.init(num_cpus=4)
 
-    args = get_args(init_args)
+    trainer = HorovodTrainer(
+        train_func,
+        train_loop_config={
+            "data_dir": "/tmp/cora",
+            "model_dir": "/tmp/model_output",
+            "sample_file": "/tmp/cora/train.nodes",
+            "partitions": 1,
```
mortid0, line 130: This is not needed. Server should be initialized outside of training loop

examples/pytorch/gat/main.py
```diff
@@ -133,17 +118,46 @@ def transform_batch(idx: list) -> dict:
...
     setup_default_logging_config(enable_telemetry=True)
-
     ray.init(num_cpus=4)
 
-    args = get_args(init_args)
+    trainer = HorovodTrainer(
+        train_func,
+        train_loop_config={
+            "data_dir": "/tmp/cora",
+            "model_dir": "/tmp/model_output",
```
mortid0, line 128: Remove

examples/pytorch/gat/main.py
```diff
@@ -133,17 +118,46 @@ def transform_batch(idx: list) -> dict:
...
-    # setup default logging component.
     setup_default_logging_config(enable_telemetry=True)
-
     ray.init(num_cpus=4)
 
-    args = get_args(init_args)
+    trainer = HorovodTrainer(
+        train_func,
+        train_loop_config={
+            "data_dir": "/tmp/cora",
```
mortid0, line 127: Remove, pass a client instead.

examples/pytorch/gat/main.py
```diff
@@ -133,17 +118,46 @@ def transform_batch(idx: list) -> dict:
...
+            "sample_file": "/tmp/cora/test.nodes",
+            "partitions": 1,
+            "num_epochs": 1,
+            "batch_size": 1000,
+            "feature_idx": 0,
+            "feature_dim": 1433,
+            "label_idx": 1,
+            "label_dim": 1,
+            "num_classes": 7,
+            "mode": "evaluate",
```
mortid0, line 158: Evaluate in a test or validation dataset directly in the training loop instead. See the pyg example in Ray.

examples/pytorch/gat/main.py
```diff
@@ -121,8 +106,8 @@ def transform_batch(idx: list) -> dict:
             losses.append(loss.item())
 
         steps_in_epoch_trained = 0
-        if epoch % args.save_ckpt_by_epochs == 0:
-            save_checkpoint(model, logger, epoch, step, args)
+        if epoch % 1 == 0:
+            save_checkpoint(model, logger, epoch, step, model_dir=model_dir)
```
mortid0, line 110: I don't think we need to save checkpoints in the examples.

examples/pytorch/gat/main.py
```diff
@@ -93,22 +74,26 @@ def train_func(config: Dict):
...
+    pipe = dataset.window(blocks_per_window=4).repeat(config["num_epochs"])
 
     def transform_batch(idx: list) -> dict:
         return model.q.query_training(g, np.array(idx))
 
     pipe = pipe.map_batches(transform_batch)
 
-    for epoch in range(epochs_trained, args.num_epochs):
+    for epoch, epoch_pipe in enumerate(pipe.iter_epochs()):
+        if epoch < epochs_trained:
```
mortid0, line 89: Not needed. Please, follow the pyg example in ray: training loop and then evaluation a test/validation dataset.

examples/pytorch/gat/main.py
```diff
@@ -93,22 +74,26 @@ def train_func(config: Dict):
     )
 
     address = "localhost:9999"
-    s = Server(address, args.data_dir, 0, len(args.partitions))
+    s = Server(address, config["data_dir"], 0, config["partitions"])
     g = DistributedClient([address])
```
mortid0, line 78: Pass a graph directly(either a distributed or in mem version) directly

examples/pytorch/gat/main.py
```diff
@@ -21,66 +21,47 @@
 from deepgnn import TrainMode, get_logger
```
mortid0, line 21: Please, cleanup imports. Lines 8, 21-24 are not needed.

examples/pytorch/gat/main.py
```diff
@@ -93,22 +74,26 @@ def train_func(config: Dict):
...
     def transform_batch(idx: list) -> dict:
         return model.q.query_training(g, np.array(idx))
 
     pipe = pipe.map_batches(transform_batch)
 
-    for epoch in range(epochs_trained, args.num_epochs):
+    for epoch, epoch_pipe in enumerate(pipe.iter_epochs()):
+        if epoch < epochs_trained:
+            continue
         scores = []
```
mortid0, line 91: Use numpy for scoring. You know in advance all the array sizes.

mortid0 (changes requested):
It is much easier to write from scratch rather than to rewrite an existing file.

examples/pytorch/gat/main.py
coledie: The "if epoch < epochs_trained:" is not needed? If so removed. The other line is the equivalent of for epoch in range(n_epochs) of that example.

examples/pytorch/gat/main.py
coledie: Compute metric requires a list of tensors, cannot be an ndarray or tensor

examples/pytorch/gat/main.py
```diff
@@ -92,58 +52,81 @@ def train_func(config: Dict):
         optimizer, named_parameters=model.named_parameters()
     )
 
-    address = "localhost:9999"
-    s = Server(address, args.data_dir, 0, len(args.partitions))
-    g = DistributedClient([address])
-    max_id = g.node_count(args.node_type) if args.max_id in [-1, None] else args.max_id
-    dataset = ray.data.range(max_id).repartition(max_id // args.batch_size)
-    pipe = dataset.window(blocks_per_window=4).repeat(args.num_epochs)
+    g = config["get_graph"]()
```
mortid0, line 55: Config is typically used for passing numerical values, not objects..
I suggest to pass graph directly to the training function.

examples/pytorch/gat/main.py
mortid0: Progress of a train metric is way too specialized requirement and is not needed in a simple example.

examples/pytorch/gat/main.py
```diff
@@ -92,58 +52,81 @@ def train_func(config: Dict):
...
+
+    for epoch, epoch_pipe in enumerate(pipe.iter_epochs()):
+        model.train()
+        train_scores = []
+        train_labels = []
         losses = []
-        for step, batch in enumerate(dataset):
-            if step < steps_in_epoch_trained:
-                continue
+        for step, batch in enumerate(
```
mortid0, line 76: doesn't look like step is used:
```python
for batch in range(epoch_pipe.iter_torch_batches(batch_size=batch_size)):
```

examples/pytorch/gat/main.py
```diff
@@ -92,58 +52,81 @@ def train_func(config: Dict):
...
-        scores = []
-        labels = []
+    test_dataset = ray.data.read_text(f"{config['data_dir']}/test.nodes")
+    test_dataset = test_dataset.repartition(1)
+    test_dataset = test_dataset.map_batches(transform_batch)
+    test_dataset_iter = test_dataset.repeat(config["num_epochs"]).iter_epochs()
+
+    for epoch, epoch_pipe in enumerate(pipe.iter_epochs()):
+        model.train()
+        train_scores = []
```
mortid0, line 73: I don't think it is train_scores/labels are very useful for a small dataset you are running this example on. It might be usefull for graphs with millions of nodes, but thousands should be train very fast and only per epoch metrics are valuable.

examples/pytorch/gat/main.py
```diff
@@ -92,58 +52,81 @@ def train_func(config: Dict):
...
 
 
 def _main():
-    # setup default logging component.
     setup_default_logging_config(enable_telemetry=True)
-
     ray.init(num_cpus=4)
 
-    args = get_args(init_args)
+    data_dir = tempfile.TemporaryDirectory()
```
mortid0, line 107: why do we need to create a folder and can't use the `Cora` object directly?

src/python/deepgnn/pytorch/common/utils.py
```diff
@@ -131,7 +131,7 @@ def load_checkpoint(model, logger=None, args=None, model_dir=".", world_rank=0):
     ckpt_path = ckpts[-1] if len(ckpts) > 0 else None
     if ckpt_path is not None:
         init_ckpt = torch.load(ckpt_path, map_location="cpu")
-        if args is not None and args.mode == TrainMode.TRAIN:
```
mortid0, line 134: why do you need changes in this file?

examples/pytorch/gat/main.py
coledie: It is used in the line below, model(batch).

examples/pytorch/gat/main.py
coledie: config is the only way to pass values to the training function and they all have to be serializable, it is easier to serialize the function than a GE.

examples/pytorch/gat/main.py
coledie: Ray re-initializes the GE many times so reloading a local client is slower and more memory consuming.

coledie: We need to create a folder so we can tell where the sample files are

examples/pytorch/gat/main.py
```diff
@@ -16,72 +14,34 @@
...
-        model.train()
-    else:
-        model.eval()
-
-    epochs_trained, steps_in_epoch_trained = load_checkpoint(
-        model, logger, args, session.get_world_rank()
-    )
+    model = train.torch.prepare_model(model)
 
     optimizer = torch.optim.Adam(
```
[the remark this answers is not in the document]
coledie: Just good practice since it is a horovod trainer.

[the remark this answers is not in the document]
mortid0: I don't think you need a horovod trainer either. Please, use simple components and leave horovod for a specialized example.

examples/pytorch/gat/main.py
mortid0: hmm, `model(batch)` doesn't use `step` variable..

examples/pytorch/gat/main.py
mortid0: Cora already has a `data_dir()` method..

examples/pytorch/gat/main.py
mortid0: why can't we pass [a dataset](https://github.com/ray-project/ray/blob/02ca4c9f2b2311e3fdf47de99303f7473170e477/python/ray/train/torch/torch_trainer.py#L250) for example?

examples/pytorch/gat/main.py
```diff
@@ -1,13 +1,11 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-from typing import Optional, Dict
+from typing import Dict
 import os
```
mortid0, line 5: not needed

examples/pytorch/gat/main.py
```diff
@@ -92,58 +52,76 @@ def train_func(config: Dict):
...
-        if epoch % args.save_ckpt_by_epochs == 0:
-            save_checkpoint(model, logger, epoch, step, args)
+        model.eval()
+        batch = next(next(test_dataset_iter).iter_torch_batches(batch_size=1000))
+        loss, score, label = model(batch)
+        test_scores = [score]
+        test_labels = [label]
 
         session.report(
             {
```
mortid0, line 91: you probably want to put `epoch` for tracking as well..

examples/pytorch/gat/main.py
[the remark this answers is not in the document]
coledie: GAT is the specialized example // always has been. It will just be this one.

examples/pytorch/gat/main.py
coledie: This is done automatically.

examples/pytorch/gat/main.py
coledie: We can but we'd have to move a lot to the main function,
```
    g = config["get_graph"]()
    batch_size = 140
    train_dataset = ray.data.read_text(f"{config['data_dir']}/train.nodes")
    train_dataset = train_dataset.repartition(train_dataset.count() // batch_size)
    train_pipe = train_dataset.window(blocks_per_window=4).repeat(config["num_epochs"])

    def transform_batch(idx: list) -> dict:
        return model.q.query_training(g, np.array(idx))

    train_pipe = train_pipe.map_batches(transform_batch)

    test_dataset = ray.data.read_text(f"{config['data_dir']}/test.nodes")
    test_dataset = test_dataset.repartition(1)
    test_dataset = test_dataset.map_batches(transform_batch)
    test_dataset_iter = test_dataset.repeat(config["num_epochs"]).iter_epochs()
```

I can just make the graph serializeable it wont affect anything though?

examples/pytorch/gat/main.py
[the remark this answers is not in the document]
mortid0: Please, pick a simple example instead of a specialized and show a typical use case.

examples/pytorch/gat/main.py
mortid0: Can we consume a graph client as a ray dataset and then apply pipeline transformations on top of it?

coledie: The problem is, the graph does not get used until the very last step, so no.

examples/pytorch/gat/main.py
mortid0: To quote the docs: `It does not necessarily map to one epoch.`
Please update a for loop then:
```python
for epoch, epoch_pipe in enumerate(train_pipe.iter_epochs()):
    # to:
for epoch_pipe in train_pipe.iter_epochs():
```

coledie: It does if you set the pipeline up how I did, but I will make this change.

examples/pytorch/gat/main.py
mortid0: why is it a problem?

coledie: I'm just not sure the idea you are talking about would look like.

mortid0: https://docs.ray.io/en/latest/ray-air/check-ingest.html#how-do-i-pass-in-a-datasetpipeline-to-my-trainer

coledie: No the "consume graph client as ray dataset" part.

mortid0: Well, you can use [map_batches operation](https://docs.ray.io/en/latest/data/api/dataset.html#ray.data.Dataset.map_batches). In that case query is the transform function and graph client is a constructor argument to the dataset. Will it work?

examples/pytorch/gat/main.py
```diff
@@ -16,72 +13,34 @@
...
-
-def train_func(config: Dict):
+def train_func(config: dict):
     """Training loop for ray trainer."""
-    args = config["args"]
-
-    logger = get_logger()
-    os.makedirs(args.save_path, exist_ok=True)
-
     hvd.init()
```
mortid0, line 24: Lets drop horovod from the simples example.

examples/pytorch/gat/main.py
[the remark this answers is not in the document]
mortid0: So which example did you pick instead?

examples/pytorch/gat/main.py
```diff
@@ -92,58 +51,72 @@ def train_func(config: Dict):
...
+        model.eval()
+        batch = next(next(test_dataset_iter).iter_torch_batches(batch_size=1000))
+        loss, score, label = model(batch)
+        test_scores = [score]
+        test_labels = [label]
 
         session.report(
             {
-                "metric": model.compute_metric(scores, labels).item(),
+                "test_metric": model.compute_metric(test_scores, test_labels).item(),
```
mortid0, line 89: replace "test_metric" with "model.metric_name()"?

examples/pytorch/gat/main.py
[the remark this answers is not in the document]
coledie: GraphSAGE.

examples/pytorch/gat/main.py
coledie: I can maybe do, but it might mess with the serialization of the graph
```
model.graph = config["get_graph"]()
train_pipe = train_pipe.map_batches(model.q.query_training)
```

examples/pytorch/gat/main.py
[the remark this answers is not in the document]
mortid0: Please, keep a single model in this PR, before it is finalized then.

examples/pytorch/gat/main.py
mortid0: > it might mess with the serialization of the graph

How? graph clients are serializable

coledie: This is what it will look like overall, good?
```
    model.g = config["get_graph"]()
    batch_size = 140
    train_dataset = ray.data.read_text(f"{config['data_dir']}/train.nodes")
    train_dataset = train_dataset.repartition(train_dataset.count() // batch_size)
    train_pipe = train_dataset.window(blocks_per_window=4).repeat(config["num_epochs"])

    train_pipe = train_pipe.map_batches(model.q.query_training)

    test_dataset = ray.data.read_text(f"{config['data_dir']}/test.nodes")
    test_dataset = test_dataset.repartition(1)
    test_dataset = test_dataset.map_batches(transform_batch)
    test_dataset_iter = test_dataset.repeat(config["num_epochs"]).iter_epochs()
```

examples/pytorch/gat/main.py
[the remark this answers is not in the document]
coledie: It looks like it is finalized except one comment and one nitpick?

examples/pytorch/gat/main.py
mortid0: No, `model.g = config["get_graph"]()` graph is a parameter to a query, not a field for a model.. So it has to be an input too a dataset a model is trained on, not the graph, like I said above: 
`graph client is a constructor argument to the dataset.`

mortid0: Sure, that works:
```python
train_pipe = train_pipe.map_batches(lambda idx: model.q.query_training(g, idx))
```

examples/pytorch/gat/main.py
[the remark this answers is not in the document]
mortid0: ok

mortid0 (changes requested):
Sage example should approximately look like this:
```python
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.
import torch
import numpy as np
import numpy as np
import ray
import ray.train as train
from ray.train.torch import TorchTrainer
from ray.air import session
from ray.air.config import ScalingConfig, RunConfig

from deepgnn.graph_engine.snark.distributed import Server, Client as DistributedClient
from deepgnn.graph_engine.data.citation import Cora
from deepgnn import setup_default_logging_config, get_logger
from deepgnn.pytorch.common import F1Score
from deepgnn.pytorch.modeling import BaseModel
from model import PTGSupervisedGraphSage  # type: ignore


def create_dataset(
    config: dict,
    model: BaseModel,
) -> ray.data.DatasetPipeline:
    g = DistributedClient(config["ge_address"])
    max_id = g.node_count(0)
    dataset = ray.data.range(max_id).repartition(max_id // config["batch_size"])
    pipe = dataset.window(blocks_per_window=4).repeat(config["num_epochs"])
    return pipe.map_batches(lambda idx: model.query(g, np.array(idx)))


def train_func(config: dict):
    """Training loop for ray trainer."""
    train.torch.accelerate()
    train.torch.enable_reproducibility(seed=session.get_world_rank())

    model = PTGSupervisedGraphSage(
        num_classes=config["label_dim"],
        metric=F1Score(),
        label_idx=config["label_idx"],
        label_dim=config["label_dim"],
        feature_dim=config["feature_dim"],
        feature_idx=config["feature_idx"],
        feature_type=np.float32,
        edge_type=0,
        fanouts=[5, 5],
        feature_enc=None,
    )
    model = train.torch.prepare_model(model)
    model.train()

    optimizer = torch.optim.SGD(
        filter(lambda p: p.requires_grad, model.parameters()),
        lr=config["learning_rate"] * session.get_world_size(),
    )
    optimizer = train.torch.prepare_optimizer(optimizer)
    dataset = config["init_dataset_fn"](
        config,
        model,
    )
    for i, epoch in enumerate(dataset.iter_epochs(max_epoch=config["num_epochs"])):
        scores = []
        labels = []
        losses = []
        for batch in epoch.iter_torch_batches(batch_size=config["batch_size"]):
            loss, score, label = model(batch)
            optimizer.zero_grad()
            loss.backward()
            optimizer.step()

            scores.append(score)
            labels.append(label)
            losses.append(loss.item())
        if i % 10 == 0:
            get_logger().info(
                f"Epoch {i:0>3d} {model.metric_name()}: {model.compute_metric(scores, labels).item():.4f} Loss: {np.mean(losses):.4f}"
            )


def _main():
    setup_default_logging_config(enable_telemetry=True)
    ray.init(num_cpus=4)

    address = "localhost:9999"
    cora = Cora()
    s = Server(address, cora.data_dir(), 0, 1)
    training_loop_config = {
        "ge_address": address,
        "batch_size": 140,
        "num_epochs": 200,
        "feature_idx": 1,
        "feature_dim": 50,
        "label_idx": 0,
        "label_dim": 121,
        "num_classes": 7,
        "learning_rate": 0.005,
        "init_dataset_fn": create_dataset,
    }

    TorchTrainer(
        train_func,
        train_loop_config=training_loop_config,
        scaling_config=ScalingConfig(num_workers=1, use_gpu=False),
        run_config=RunConfig(
            verbose=0,
        ),
    ).fit()


if __name__ == "__main__":
    _main()
```
Please, improve it to make it verifiable in our CI pipeline.

examples/pytorch/graphsage/main.py
```diff
@@ -135,12 +122,13 @@ def train_func(config: Dict):
 
         losses_full.extend(losses)
         steps_in_epoch_trained = 0
-        if epoch % args.save_ckpt_by_epochs == 0:
-            save_checkpoint(model, logger, epoch, step, args)
+
+        if "model_path" in config:
+            save_checkpoint(model, logger, epoch, step, model_dir=config["model_path"])
```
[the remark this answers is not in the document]
coledie: I've tried this before and it is not easy to work with, it was better to just use the utility functions. I forget what the issues were exactly.

coledie:
This is much better but the reason I cannot remove the weird dataset stuff and the checkpoint fn, ... is because of we still need to support the unit test.

examples/pytorch/graphsage/main.py
```diff
@@ -39,76 +40,58 @@ def init_args(parser: argparse.Namespace):
...
-        args,
+        config,
         model,
         rank=session.get_world_rank(),
         world_size=session.get_world_size(),
-        address=address,
+        get_graph=config["get_graph"],
     )
     losses_full = []
     epoch_iter = (
```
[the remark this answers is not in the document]
coledie: The unit test support that uses mockdataloader.

examples/pytorch/graphsage/main.py
```diff
@@ -135,12 +122,13 @@ def train_func(config: Dict):
...
-        if epoch % args.save_ckpt_by_epochs == 0:
-            save_checkpoint(model, logger, epoch, step, args)
+
+        if "model_path" in config:
+            save_checkpoint(model, logger, epoch, step, model_dir=config["model_path"])
 
         session.report(
             {
-                "metric": model.compute_metric(scores, labels).item(),
+                model.metric_name(): model.compute_metric(scores, labels).item(),
```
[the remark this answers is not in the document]
coledie: Need to track historical losses for unit tests.

examples/pytorch/graphsage/main.py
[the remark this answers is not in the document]
mortid0: what component are you trying to unit test?

examples/pytorch/graphsage/main.py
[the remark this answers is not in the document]
mortid0: why console doesn't work for tracking?

examples/pytorch/graphsage/main.py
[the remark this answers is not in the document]
mortid0: Please, write explicit issues you had with it. I don't see any new component in the main.py and don't understand what exactly we need to unit test.

examples/pytorch/graphsage/main.py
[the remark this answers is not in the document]
coledie: The GraphSAGE model, you said to keep the mockdataloader in previous PR.

examples/pytorch/graphsage/main.py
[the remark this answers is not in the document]
coledie: Can we remove the unit tests for graphsage?

examples/pytorch/graphsage/main.py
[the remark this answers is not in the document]
mortid0: This is a `main.py` file, not `model.py`. Any changes in the former shouldn't prevent you from testing the latter.

examples/pytorch/graphsage/main.py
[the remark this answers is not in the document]
mortid0: You are mixing two separate components in one:
1. Model.py still should be unit tested
2. We still need to run main.py in CI pipeline and check the output makes sense. Something like:
```bash
python main.py > log.txt
if ["($grep -ic 'Epoch 100: F1Score: 0.1' log.txt )" != "3"]
  then
    exit 3
fi
```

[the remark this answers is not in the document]
coledie: I did not write the unit test, should I update it so that it only tests model.py? Ideally we just have main.py since if that works then the model works.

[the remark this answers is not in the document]
mortid0: > should I update it so that it only tests model.py?

yes

> main.py since if that works then the model works.

we also want to make sure if anyone runs examples, they work as expected.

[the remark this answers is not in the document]
coledie: So to be clear - right now the unit test tests both the model and main.py by using the training loop in main.py and checking accuracy - but you want me to switch it so there are separate training loops in the unit test and main.py and we will test them separately?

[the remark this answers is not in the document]
mortid0: > So to be clear - right now the unit test tests both the model and main.py by using the training loop in main.py and checking accuracy

You are not using old data loaders anymore, so there is no point of sticking to testing model twice.

> but you want me to switch it so there are separate training loops in the unit test and main.py and we will test them separately?

You don't need a training loop in a unit test at all. Update it to check model queries graph/generates embeddings correctly, that's it. E2E can be verified in console.

> This seems like a lot of duplicate code and hacky unit test just to save a few lines in the example.

There will be way less code with this aproach.

examples/pytorch/gat/main.py
```diff
@@ -92,58 +51,74 @@ def train_func(config: Dict):
...
-    # setup default logging component.
     setup_default_logging_config(enable_telemetry=True)
-
     ray.init(num_cpus=4)
 
-    args = get_args(init_args)
+    cora = Cora()
+
+    address = "localhost:9999"
+    s = Server(address, cora.data_dir(), 0, 1)
```
coledie, line 104: Potentially move inside fo train_func to support multi machine users

examples/pytorch/gat/output_check.sh
```diff
@@ -0,0 +1,4 @@
+export SNARK_LIB_PATH="$(pwd)/src/cc/lib/libwrapper.so"
+python examples/pytorch/gat/main.py > log.txt
+echo "$(cat log.txt)"
+if [ "$(grep -ic '0.826 | 0.721931' log.txt)" != "1" ]; then exit 3; fi
```
mortid0, line 4: instead of copy pasting this file in every model:
1. extract common functionality
2. pass expected values as command line arguments.

docs/torch/link_pred.rst
```diff
@@ -312,7 +312,7 @@ Finally we train the model to predict whether an edge exists between any two nod
 
     >>> model_dir = tempfile.TemporaryDirectory()
 
-    >>> ray.init(num_cpus=3)
+    >>> ray.init(num_cpus=4)
```
mortid0, line 315: why do we need to increase number of cpus?

examples/pytorch/graphsage/main.py
```diff
@@ -133,48 +60,35 @@ def train_func(config: Dict):
             labels.append(label)
             losses.append(loss.item())
 
-        losses_full.extend(losses)
-        steps_in_epoch_trained = 0
-        if epoch % args.save_ckpt_by_epochs == 0:
-            save_checkpoint(model, logger, epoch, step, args)
+        if i % 10 == 0:
+            print(
```
mortid0, line 64: use logger for output.

coledie: It needs to be print so it shows up in the log.txt in the test script.

PR #157, opened 2023-02-07
Add ppr-go neighbor sampling strategy

mortid0:

New Behavior
----------------
Added new feature: personalized page rank sampling. We use https://github.com/TUM-DAML/pprgo_pytorch implementation, but in C++ instead of python.

### DIFF
--- a/src/cc/lib/py_graph.cc
+++ b/src/cc/lib/py_graph.cc
@@ -926,6 +926,60 @@ int32_t RandomWalk(PyGraph *py_graph, int64_t seed, float p, float q, NodeID def
     return 0;
 }
 
+namespace
+{
+
+// Auxilarry data structure to group all variables in one place.
+typedef struct
+{
+    absl::flat_hash_map<NodeID, size_t> lookup;
+    std::vector<NodeID> node_ids;
+    std::vector<size_t> nb_index;
+    std::vector<uint64_t> counts;
+} NB_Count_Cache;
+
+// Cache neighbor counts for each node in the input list.
+void lookup_neighbor_counts(PyGraph *py_graph, NB_Count_Cache &cache, const std::vector<NodeID> &neighbors,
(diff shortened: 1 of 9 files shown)


src/cc/lib/py_graph.cc
```diff
@@ -925,6 +926,123 @@ int32_t RandomWalk(PyGraph *py_graph, int64_t seed, float p, float q, NodeID def
...
+            }
+            else
+            {
+                const auto &top = pq.top();
+                if (top.first > kv.second)
+                {
+                    continue;
+                }
+                pq.pop();
+                pq.emplace(kv.second, kv.first);
```
coledie, line 1020: Should we add this to benchmarks? Does it scale?

src/python/deepgnn/graph_engine/snark/tests/neighbor_sampler_test.py
```diff
@@ -596,6 +595,78 @@ def test_karate_club_weighted_neighbor_sampling_different_result(
...
+        edge_types=12,
+        count=2,
+        alpha=0.2,
+        eps=0.0001,
+        default_node=-1,
+        default_weight=0.0,
+    )
+
+    npt.assert_array_equal(nodes, [[1, -1], [2, -1], [6, -1]])
+    npt.assert_array_almost_equal(
```
coledie, line 636: is this [src, dst], if so why is it -1 for dst always?

src/cc/lib/py_graph.cc
mortid0: PPR sampling is similar to random walk. It is implemented at the library interface level and we can't include it in existing benchmarks. Scaling similarly depends on input parameters, alpha and eps.

src/python/deepgnn/graph_engine/snark/tests/neighbor_sampler_test.py
mortid0: Because there are no neighbors with type 12..

coledie (approved):

PR #161, opened 2023-02-09
Pick a single sparse feature from multiple servers.

mortid0:

Previous Behavior
----------------
Sparse feature values were corrupted if it is present on multiple GE servers(with different values).

New Behavior
----------------
It is hard to control duplicate features across multiple GE server instances, we can pick a random value for sparse features.
In case of dense/string features, they are already overwritten, but for sparse features we were merging indices concurrently.
We pick a random source during feature extraction and stick to it until values returned to python.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Implement del method to release C++ client and server. Important for ray actors, because they create numerous clients during training.
 
+- If sparse feature values present on multiple servers, then only one will be returned with source picked randomly.
+
 ## [0.1.57] - 2022-12-15
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Implement del method to release C++ client and server. Important for ray actors, because they create numerous clients during training.
 
+- If sparse feature values present on multiple servers, then only one will be returned with source picked randomly.
```
coledie, line 21: Isn't it a user error if they load feature values to multiple servers with the same index?

mortid0: yes, but it is hard for them to fix it, so we can do our best to resolve this issue.

coledie (approved):

Issue #160, opened 2023-02-08
Connection reset by peer retry 


MortezaRamezani:
- [ ] Issue is labeled using the label menu on the right side.

Environment
-----------
* Python version: `Python 3.7.10`
* deepgnn-ge Version: `0.1.58.dev5`
* deepgnn-torch Version: N/A
* deepgnn-tf Version: `0.1.58.dev5`
* OS: CentOS Linux release 7.9.2009, 3.10.0

Issue Details
-----------
* Download a dataset(`ppi`)
`python -m deepgnn.graph_engine.data.ppi --data_dir /tmp/ppi`
* Create a client/server as follows:
```python
# client.py
import json
import numpy as np
from deepgnn.graph_engine.snark.client import DistributedGraph

service_config_json = json.dumps({
                                    "methodConfig": [
                                        {
                                            "name": [{}],
                                            "waitForReady": True,
                                            "retryPolicy": {
                                                    "maxAttempts": 5,
                                                    "initialBackoff": "2s",
                                                    "maxBackoff": "1s",
                                                    "backoffMultiplier": 2,
                                                    "retryableStatusCodes": ["UNAVAILABLE", "ABORTED"],
                                                },

                                    }]
                                })
options = []
options.append(("grpc.enable_retries", 1))
options.append(("grpc.service_config", service_config_json))
g = DistributedGraph(["0.0.0.0:9090"], grpc_options=options)
for i in range(10000):
    f = g.node_features(np.array([0, 1, 13], dtype=np.int64), [[0, 100000000]], dtype=np.float32)
    print(f"#{i} features {f.shape}")
```

```python
# server.py
from deepgnn.graph_engine.snark.server import Server
s = Server("/tmp/ppi", [0], "0.0.0.0:9090")
input("Server started..")
```
* Use `tcpkill` to interrupt the connection
`sudo tcpkill -i lo -9 port 9090`

* Expected behavior
```
E00000000 00:00:00.000000  1473 client.cc:309] RAW: Request failed, code: 14. Message: recvmsg:Connection reset by peer
E00000000 00:00:00.000000  1431 py_graph.cc:385] RAW: Exception while fetching node features: Request failed. Message: recvmsg:Connection reset by peer
Traceback (most recent call last):
  File "client.py", line 27, in <module>
    f = g.node_features(np.array([0, 1, 13], dtype=np.int64), [[0, 100000000]], dtype=np.float32)
[...]

PR #163, opened 2023-02-10
add retry to snark client

MortezaRamezani:
- [ ] Changelog and documentation updated.

Previous Behavior
----------------
Various connection issues can lead to unexpected code behavior #160 

New Behavior
----------------
The snark client will retry on failed queries using the already defined `grpc_options`.

Notes
----------------
This problem is occurring due to grpc design, where retry is not valid on [_committed RPCs_](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#when-retries-are-valid). Current solution is not the most efficient way of solving it as it retry the whole query in case of any failure. The better way might be to handle failed connection inside `client.cc` to avoid redundant calls to healthy servers.

### DIFF
--- a/src/python/deepgnn/graph_engine/snark/client.py
+++ b/src/python/deepgnn/graph_engine/snark/client.py
@@ -25,43 +25,51 @@ from ctypes import (
 from typing import Any, List, Tuple, Union, Optional, Sequence, Dict
 from enum import IntEnum
 
-import time
 import json
 import numpy as np
-
+from tenacity import Retrying, stop_after_attempt, wait_exponential
+from deepgnn import get_logger
 from deepgnn.graph_engine.snark._lib import _get_c_lib
 from deepgnn.graph_engine.snark._downloader import download_graph_data, GraphPath
 from deepgnn.graph_engine.snark.meta import Meta
 
 
-def _parse_grpc_options(options: List[tuple]) -> Dict:
-    """Parsing the grpc_options to use on client side retry calls"""
+def _parse_grpc_options(options: List[Tuple[str, str]]) -> Dict:
+    """Parse the grpc_options to use on client side retry calls."""
     for opt in options:
         if opt[0] == "grpc.service_config":
             try:
                 config = json.loads(opt[1])
-                retries = int(config['methodConfig'][0]['retryPolicy']['maxAttempts'])
-                delay_str = config['methodConfig'][0]['retryPolicy']['initialBackoff']
-                delay = datetime.strptime(delay_str, '%Ss').second
-                return {"retries": retries, "delay": delay}
-            except Exception as e:
-                print('Cannot parse the grpc_options, set the deafult value')
+                max_attempts = int(
+                    config["methodConfig"][0]["retryPolicy"]["maxAttempts"]
+                )
+                initial_backoff = datetime.strptime(
+                    config["methodConfig"][0]["retryPolicy"]["initialBackoff"], "%Ss"
+                ).second
+                max_backoff = datetime.strptime(
+                    config["methodConfig"][0]["retryPolicy"]["maxBackoff"], "%Ss"
+                ).second
+                backoff_multiplier = int(
+                    config["methodConfig"][0]["retryPolicy"]["backoffMultiplier"]
+                )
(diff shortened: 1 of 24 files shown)


MortezaRamezani:
@microsoft-github-policy-service agree company="Linkedin"

src/python/deepgnn/graph_engine/snark/client.py
```diff
@@ -224,6 +256,10 @@ def __init__(
         )
         self._describe_clib_functions()
 
+        self.dist_mode = False # used for DistGraph only
+        self.retries = None # used for DistGraph only
```
mortid0, line 260: Can you use a retry library instead? I suggest tenacity: https://github.com/jd/tenacity
You can have a custom retryer field that can wrap existing functions and implements jitter and exponential backoff. It will avoid code duplication then. Something like this:
```python
class MemoryGraph:
    def __init__ (self, ...):
        self._retryer = Retrying(stop=stop_after_attempt(max_attempts=0), reraise=True)
        ...
        def node_features(...):
            ....
            self._retryer(self.lib.GetNodeFeature, ...)
            
class DistributedGraph:
    def __init__ (self, ...):
        retry_options = _parse_grpc_options(...)
        self._retryer = Retrying(stop=stop_after_attempt(max_attempts=retry_options['retries']), reraise=True)
```

src/python/deepgnn/graph_engine/snark/client.py
```diff
@@ -22,16 +22,48 @@
...
+    for opt in options:
+        if opt[0] == "grpc.service_config":
+            try:
+                config = json.loads(opt[1])
+                retries = int(config['methodConfig'][0]['retryPolicy']['maxAttempts'])
+                delay_str = config['methodConfig'][0]['retryPolicy']['initialBackoff']
+                delay = datetime.strptime(delay_str, '%Ss').second
+                return {"retries": retries, "delay": delay}
+            except Exception as e:
+                print('Cannot parse the grpc_options, set the deafult value')
```
mortid0, line 48: Please, use `get_logger` for logging errors.

src/python/deepgnn/graph_engine/snark/client.py
```diff
@@ -22,16 +22,48 @@
...
+    i = 0
+    while True:
+        try:
+            return func(*args)
+        except Exception as ex:
+            if i >= retries:
+                raise RuntimeError(f"Got exception in {func.__name__}: {str(ex)}")
+            i += 1
+            time.sleep(delay)
+            print(f'Trying {func.__name__} again, retry #{i+1}...')
```
mortid0, line 64: Same as above, use logger.

src/python/deepgnn/graph_engine/snark/client.py
MortezaRamezani: So we have retry for MemoryGraph (non-distributed) as well?

mortid0: Well, it will be a no-op retry.

mortid0 (commented):
Overall looks good! Please address nits and we'll merge it.

mortid0:
* for failing tests: update BUILD files and include tenacity as dependency.
* include `tenacity>=8` in [setup.py](https://github.com/microsoft/DeepGNN/blob/cd2497aae41ee22d6c30dbcd2f6d53de3c36d193/src/python/setup.py#L85)

MortezaRamezani:
> 

Thanks for the suggestions. General question, I have difficulties installing all the packages in requirements.txt in `ubuntu22.04` and `python3.10` images. Is that expected?

mortid0:
> General question, I have difficulties installing all the packages in requirements.txt in ubuntu22.04 and python3.10 images. Is that expected?

You don't need to install these packages at all, bazel will take care of it for test targets via BUILD files. All you need to run is:
```
bazel test //... -c dbg --config=linux --test_output==all
```

mortid0 (approved):

MortezaRamezani:
The test failed after merge. Is it related to #159 ?

mortid0:
There is another unrelated build issue. A retry should help.

PR #179, opened 2023-05-10
Add Ray on AML multi node example.

coledie:

New Behavior
----------------
* Will merge into Ray PR w/ Ray model code.
* Provides example using multi node Ray on AML cluster setup with GEs connecting across nodes and synchronizing.


examples/pytorch/aml.py
```diff
@@ -0,0 +1,152 @@
...
+        sleep(1)
+    address = ray.get(
+        [global_cache.get.remote(i) for i in range(session.get_world_size())]
+    )
+
+    _ = Server(
+        address[session.get_world_rank()], ppi.data_dir(), session.get_world_size(), 1
+    )
+
+    cl = DistributedClient(address)
```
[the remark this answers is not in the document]
coledie: set_lock will pause the program until all servers are ready, so by the time it gets to this line, all servers will be ready.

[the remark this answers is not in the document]
coledie: Which part?

examples/pytorch/aml.py
```diff
@@ -0,0 +1,152 @@
...
+from deepgnn.graph_engine import Graph, graph_ops
+from deepgnn.pytorch.modeling import BaseModel
+from deepgnn.graph_engine.data.ppi import PPI
+from deepgnn.graph_engine.snark.distributed import Server, Client as DistributedClient
+from time import sleep
+import socket
+
+
+@ray.remote
+class Cache:
```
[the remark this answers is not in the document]
coledie: Will keep in example for this PR.

examples/pytorch/aml.py
```diff
@@ -0,0 +1,156 @@
...
+        [server_lock.get.remote(i) for i in range(session.get_world_size())]
+    )
+
+    Server(
+        address[session.get_world_rank()], ppi.data_dir(), session.get_world_size(), 1
+    )
+
+    cl = DistributedClient(address)
+
+    # TODO Replace these 2 lines with a model
```
mortid0, line 93: use an assert to extract some node features from graph to confirm everything works as expected.

examples/pytorch/aml.py
```diff
@@ -0,0 +1,156 @@
...
+def release_lock(server_lock, rank):
+    """Release lock at rank, then halt until all servers released."""
+    server_lock.release_lock.remote(session.get_world_rank())
+    while not ray.get(server_lock.wait_released.remote()):
+        sleep(1)
+
+
+def train_func(config: dict):
+    """Training loop for ray trainer."""
+    ppi = PPI(num_partitions=session.get_world_size())
```
mortid0, line 75: Please, use a smaller graph, like cora to reduce time to run this example.

examples/pytorch/aml.py
```diff
@@ -0,0 +1,156 @@
...
+    release_lock(server_lock, session.get_world_rank())
+
+
+if __name__ == "__main__":
+    import azureml
+    from azureml.core import Workspace
+    from ray_on_aml.core import Ray_On_AML
+
+    aml = True
+    try:
```
mortid0, line 106: This try block is more confusing than helpful. I'd suggest to comment it out and add instructions how to run it on AML.

examples/pytorch/aml.py
```diff
@@ -0,0 +1,156 @@
...
+        ray_on_aml = type("Ray_On_AML", (object,), {"shutdown": (lambda: None)})
+
+    try:
+        num_workers = 2
+        server_lock = ServerLock.remote(num_workers)  # type: ignore
+
+        trainer = TorchTrainer(
+            train_func,
+            train_loop_config={
+                "num_epochs": 10,
```
mortid0, line 139: These parameters are not used in the train loop.

examples/pytorch/aml.py
```diff
@@ -0,0 +1,156 @@
...
+        f"{ray._private.services.get_node_ip_address()}:999{session.get_world_rank()}"
+    )
+
+    set_lock(server_lock, session.get_world_rank(), address)
+
+    address = ray.get(
+        [server_lock.get.remote(i) for i in range(session.get_world_size())]
+    )
+
+    Server(
```
mortid0, line 87: 1. Why do we need locks if the server is created before client in the training loop?
2. Where do you shutdown server?
3. Could you separate "business" logic of using a graph for querying with infrastructure logic of setting up multiple servers and connecting a client to them when they are ready.

coledie: 1. Locks are since a server is created in each worker/node so they will not necessarily be created before every client.
2. We wait for all locks to be released at the end, then they go out of scope and get closed.
3. Could you clarify // show code for this one.

examples/pytorch/aml.py
coledie: I'll submit another PR to add the num_partitions parmaeter to cora then do it.

examples/pytorch/aml.py
coledie: Will do after cora.

examples/pytorch/aml.py
mortid0: 1. Locks should have a wait before a client is created, because other workers might start late and a client will fail to connect to the server located on it.
2. Thanks!
3. You probably want to have something like this: https://docs.ray.io/en/latest/ray-core/patterns/tree-of-actors.html A supervisor starts GE servers in a constructor and then calls fit for every worker it is responsible for in a separate method.

coledie: 1. Updated.
3. What is the benefit of this added complexity? My simple solution seems to work.

examples/pytorch/aml.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@ray.remote
+class ServerLock:
+    """Lock for server."""
+
+    def __init__(self, size):
+        """Initialize server lock."""
+        self.cache = {}
+        self.locks = [0 for _ in range(size)]
```
coledie, line 33: Move classes to separate file

examples/pytorch/aml.py
```diff
@@ -0,0 +1,162 @@
...
+
+    with ServerContext(
+        session.get_world_size(), session.get_world_rank(), address
+    ) as lock:
+        cl = DistributedClient(lock.get_address())
+
+        features = cl.node_features(np.array([0, 1]), np.array([[0, 1]]), np.float32)
+        npt.assert_equal(features, np.array([[0.0], [0.0]]))
+
+        sleep(5)
```
mortid0, line 110: Please, remove sleep everywhere.

examples/pytorch/aml.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Server(address, cora.data_dir(), session.get_world_size(), 1)
+
+    with ServerContext(
+        session.get_world_size(), session.get_world_rank(), address
+    ) as lock:
+        cl = DistributedClient(lock.get_address())
+
+        features = cl.node_features(np.array([0, 1]), np.array([[0, 1]]), np.float32)
+        npt.assert_equal(features, np.array([[0.0], [0.0]]))
```
mortid0, line 108: we backfill with 0s if a feature is not found too. Could you add some non-default values(probably with feature_id 1, instead of 0)?

coledie: For some reason, all node features for index 0 and 1 have no features up to length 121, even for dtype int64

mortid0: other node_ids might work too..

examples/pytorch/aml.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+if __name__ == "__main__":
+    import sys
+    from azureml.core import Workspace
+    from ray_on_aml.core import Ray_On_AML
+
+    aml = True
+    if len(sys.argv) < 2 or sys.argv[1] != "--unit_test":
+        ws = Workspace.from_config()
```
mortid0, line 116: why do we need lines 116-135 in this file?

coledie: This is so it will actually run for users in multi node AML

examples/pytorch/aml.py
coledie: It looks like this is a bug somehow, cora.features works but cl.features doesn't ever

examples/pytorch/aml.py
mortid0: How do we know it will keep working if it is never executed in CI?

coledie: It is simple code from ray_on_aml which is heavily tested by them, so we can count on it

mortid0 (approved):

PR #180, opened 2023-05-17
Implement uniform sampling for temporal graphs and add timestamps to result tuples

mortid0:

Previous Behavior
----------------
We return 4 `np.arrays` in the `sample_neighborw` method from `_base.py`, but didn't fill the last element in that tuple.

New Behavior
----------------
1. Add timestamp information about the edge connecting 2 nodes, when it was created.
2. Implement uniform sampling for temporal graphs. I though it was added with temporal graphs, but somehow missed it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.1.60] - 2022-04-18
 
 ### Added
+- Every sampling method returns created_at timestamps as the last argument.
+
+## [0.1.60] - 2022-04-18
+
+### Added
 - Breaking. Temporal graph support. Custom decoders must add 2 optional integers in returned tuple in  `decode` method, representing `created_at` and `removed_at` fields. Metadata file must have a `watermark` field.
 
 - Last N created neighbors sampling method for temporal graphs.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.1.60] - 2022-04-18
 
 ### Added
-- Every sampling method returns created_at timestamps as the last argument.
+- Add `return_edge_created_ts` argument to neighbor sampling methods to return timestamps when edges connecting nodes were created.
+
+### Fixed
+- Uniform sampling works in temporal graphs.
 
 ## [0.1.60] - 2022-04-18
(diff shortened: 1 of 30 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Every sampling method returns created_at timestamps as the last argument.
```
coledie, line 10: In the docstrings it says timestamps are only returned for temporal graphs but here it looks like they are always returned.

coledie (approved):

CHANGELOG.md
mortid0: I need to update docstrings, thank!

coledie: Can you add a parameter to enable returning timestamps, I believe this is standard practice instead of returning many people info they do not need: https://numpy.org/doc/stable/reference/generated/numpy.unique.html

mortid0: Make sense, Let me add it.

mortid0: Fixed.

src/python/deepgnn/graph_engine/snark/local.py
```diff
@@ -118,8 +122,16 @@ def sample_neighbors(
                 default_weight,
                 seed=int(random.getrandbits(64)),
                 timestamps=timestamps,
+                return_edge_created_ts=return_edge_created_ts,
+            )
+            if return_edge_created_ts:
```
coledie, line 127: Should be if not return_edge_created_ts?

coledie (approved):

src/python/deepgnn/graph_engine/snark/local.py
mortid0: No. If `return_edge_created_ts` is set to `True` we want to return everything, because tuple will have 4 elements. Otherwise, the tuple will have only 3 elements and we need to backfill.

src/python/deepgnn/graph_engine/snark/local.py
```diff
@@ -118,8 +122,16 @@ def sample_neighbors(
...
+                return result
+            return (
+                result[0],
+                result[1],
+                result[2],
+                np.full(result[0].shape, -1, dtype=np.int64),
```
coledie, line 133: Why do we need this backfill? Could the return just be a 3 tuple if timestamps are not needed?

PR #194, opened 2023-07-03
Add PyG remote backend example.

coledie:

New Behavior
----------------
Add example with PyG remote backend example.


examples/pytorch/pyg/train.py
```diff
@@ -0,0 +1,158 @@
...
+    model = GraphSAGE(
+        in_channels=50,
+        hidden_channels=64,
+        num_layers=2,
+        out_channels=64,
+    ).to(device)
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.005)
+
+    for epoch in range(1, 6):
+        loss = train()
```
mortid0, line 157: Please, add a a test() method similar [to this](https://github.com/pyg-team/pytorch_geometric/blob/decec47e3c9cd67b8ddbd58a15edd00de0c530ef/examples/graph_sage_unsup.py#L57C1-L67C29)(with evaluation only on a test dataset) and an assert to make sure we get expected values from training.

examples/pytorch/pyg/test_interface.py
```diff
@@ -0,0 +1,53 @@
...
+import argparse
+import torch
+
+from examples.pytorch.pyg.train import DeepGNNFeatureStore, DeepGNNGraphStore
+from deepgnn.graph_engine import SamplingStrategy
+from deepgnn.graph_engine.data.citation import Cora
+from torch_geometric.loader import LinkNeighborLoader
+
+
+def test_interface():
```
mortid0, line 19: This is not needed, just use an assert in the train.py

examples/pytorch/pyg/BUILD
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+load("@rules_python//python:defs.bzl", "py_test", "py_library")
+load("@pip_deps//:requirements.bzl", "requirement")
+
+py_library(
```
mortid0, line 7: 1. it should be a binary, so you can run it as an example.
2. Rename train.py to main.py. This is a Bazel convention I think.

examples/pytorch/pyg/test_interface.py
coledie: Train.py uses ppi which is too slow to run in CI pipeline.

mortid0: what stops you from using a smaller graph, which will be also faster to load?

coledie: We want it to be a useful example that shows actual learning

mortid0: > We want it to be a useful example that shows actual learning. Cora is not big enough.

Why many papers use it for benchmarks?

> Also none of our other examples do testing like this

[Of course we do](https://github.com/microsoft/DeepGNN/blob/1e35f511e6f18857e1aa0c774e9dee23b9adc88c/.github/workflows/ci.yml#L154C9-L155C73). There is no assert in main, we only verify code runs without exceptions and have UT to verify logic. It will be easier to add an assert in this train.py and skip UTs, because you are not adding any new code outside of examples folder.

examples/pytorch/pyg/train.py
coledie: Is it possible with Cora since all edges have type 0? It would just be the training set accuracy which maybe is okay also?

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,160 @@
...
+        """To be implemented by :class:`GraphStore` subclasses."""
+        edge_type = int(edge_attr.edge_type[1])
+        edge = self.ge.sample_edges(
+            self.ge.edge_count(0), np.array(edge_type), SamplingStrategy.Random
+        )
+        edge_index = (
+            torch.Tensor(edge[:, 0]).long(),
+            torch.Tensor(edge[:, 1]).long(),
+        )
+        # assert False, edge_index
```
mortid0, line 89: Please, remove comments.

examples/pytorch/pyg/train.py
coledie: I added the test but even though loss is going down, test stays random, do you see anything wrong with it?

examples/pytorch/pyg/BUILD
```diff
@@ -0,0 +1,52 @@
...
+        "//src/python/deepgnn/pytorch/encoding/twinbert:deepgnn_pytorch_encoding_twinbert",
+        "//src/python/deepgnn/pytorch/modeling:deepgnn_pytorch_modeling",
+        "//src/python/deepgnn/pytorch/encoding/twinbert/deepspeed:deepgnn_pytorch_encoding_twinbert_ds",
+        "//src/python/deepgnn/pytorch/nn:deepgnn_pytorch_nn",
+        "//src/python/deepgnn/pytorch/training:deepgnn_pytorch_training",
+        "//examples/pytorch:example_torch_tests",
+    ],
+)
+
+py_test(
```
mortid0, line 29: This is not needed anymore

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,188 @@
+"""PyGeo interface training example.
+
+https://github.com/pyg-team/pytorch_geometric/blob/ae84a38f14591ba9b8ce64e704e04ea1271c3b78/examples/graph_sage_unsup_ppi.py#L12
```
mortid0, line 3: This comment doesn't seem to be relevant anymore, because you are using Cora.

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,188 @@
...
+class DeepGNNFeatureStore(FeatureStore):
+    """An abstract base class to access features from a remote feature store."""
+
+    def __init__(self, ge):
+        """Initialize DeepGNN feature store."""
+        super().__init__()
+        self.ge = ge
+
+    def _put_tensor(self, tensor, attr) -> bool:
+        """To be implemented by :class:`FeatureStore` subclasses."""
```
mortid0, line 30: This docstring is not needed. Same below.

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,188 @@
...
+        for i in ["x", "y"]:
+            ta = TensorAttr()
+            ta.group_name = "0"
+            ta.attr_name = f"{i}"
+            output.append(ta)
+        return output
+
+
+class DeepGNNGraphStore(GraphStore):
+    """An abstract base class to access edges from a remote graph store.
```
mortid0, line 59: Please update docstrings, DeepGNNGraphStore is definitely not abstract.

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,188 @@
...
+        ),
+        batch_size=140,
+        shuffle=True,
+        neg_sampling_ratio=1.0,
+        num_neighbors=[5, 5],
+        num_workers=1,
+        persistent_workers=True,
+        edge_label_index=(
+            ("0", "0", "0"),
+            torch.Tensor(ge.sample_edges(2708, np.array(0), SamplingStrategy.Random))
```
mortid0, line 153: Please, extract a const for this magic number, 2708.

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,188 @@
...
+        num_layers=2,
+        out_channels=7,
+    ).to(device)
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.005)
+
+    for epoch in range(1, 10):
+        loss = train()
+        test_f1 = test()
+        print(f"Epoch: {epoch:03d}, Loss: {loss:.4f}, Test F1: {test_f1:.4f}")
+    assert loss <= 0.55
```
mortid0, line 188: You want to assert f1 value on test dataset, not loss.

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,188 @@
...
+        hidden_channels=64,
+        num_layers=2,
+        out_channels=7,
+    ).to(device)
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.005)
+
+    for epoch in range(1, 10):
+        loss = train()
+        test_f1 = test()
+        print(f"Epoch: {epoch:03d}, Loss: {loss:.4f}, Test F1: {test_f1:.4f}")
```
mortid0, line 187: Please, use logger instead of prints

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,188 @@
...
+        optimizer.step()
+
+        total_loss += float(loss) * link_pred.numel()
+        total_examples += link_pred.numel()
+
+    return total_loss / total_examples
+
+
+def test():
+    for data in tqdm.tqdm(loader):
```
mortid0, line 130: forgot `model.eval()`

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,188 @@
...
+        """Obtain all edge attributes stored in the :class:`GraphStore`."""
+        output = []
+        ta = EdgeAttr(("0", "0", "0"), "coo", size=[self.ge.node_count(0), 2708])
+        output.append(ta)
+
+        return output
+
+
+def train():
+    """Train the model."""
```
mortid0, line 104: Self explanatory docstring. Either remove here or add to the test function below.

examples/pytorch/pyg/main.py
coledie: I am using their graphsage model still and most of their training loop.

examples/pytorch/pyg/main.py
coledie: See other comment above, I need help with this.

examples/pytorch/pyg/BUILD
coledie: The whole build file?

examples/pytorch/pyg/main.py
mortid0: Training loop is the same across many models. Please, remove this comment.

examples/pytorch/pyg/train.py
mortid0: Could you explain, what is random? f1 score stays low or non deterministic? If the former, you need to debug outputs of wrappers and compare them to relevant PyG tensors. If the latter, you need to set random seed. Either [globally](https://github.com/microsoft/DeepGNN/blob/1e35f511e6f18857e1aa0c774e9dee23b9adc88c/src/python/deepgnn/pytorch/common/utils.py#L31), or in the calls to sampling methods.

examples/pytorch/pyg/main.py
```diff
@@ -0,0 +1,196 @@
...
+            ta.attr_name = f"{i}"
+            output.append(ta)
+        return output
+
+
+class DeepGNNGraphStore(GraphStore):
+    """A class to access edges from a remote graph store.
+
+    Args:
+        edge_attr_cls (EdgeAttr, optional): A user-defined
```
mortid0, line 58: This is a wrong docstring

mortid0: There is no `edge_attr_cls` argument

examples/pytorch/pyg/main.py
mortid0: 2708 is still not extracted as const. I believe this should be number of edges of type 0 or something like that.

.github/workflows/ci.yml
```diff
@@ -145,6 +145,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install torch
         run: pip install --timeout 300 torch==1.13.1
+      - name: Pip freeze
```
mortid0, line 148: why is this needed?

mortid0 (approved):

.github/workflows/ci.yml
coledie: Debug, it looks like the CI is failing because of new PyG version, this will hopefully get resolved with Ray fix so I will finalize then.

mortid0 (approved):

PR #217, opened 2023-08-15
Produce wheels with bazel targets.

mortid0:

Previous Behavior
----------------
We use env variables to point to current version of shared library in tests. This requires copy pasting code across all tests using C++ in python.

New Behavior
----------------
Create a wheel artifact for all python tests. There is no need to override library location anymore and we can test e2e without CI scripts - directly with Bazel. We use external `rules_pycross` because `rules_python` don't implement PyInfo return interface for `py_wheel` targets.


requirements.txt
```diff
@@ -1,5 +1,5 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
--f https://data.pyg.org/whl/torch-1.12.1+cpu.html  # For torch geometric skip torch-scatter/sparse setup.py
```
mortid0, line 2: had to remove comments because new python rules have trouble parsing them :(

coledie: If pyg still works without it then thats great

coledie (approved):

PR #218, opened 2023-08-18
[CI] Setup remote cache on linux

mortid0:

Previous Behavior
----------------
CI pipeline was slow and failing with out of storage errors

New Behavior
----------------
Added [bazel-remote](https://github.com/buchgr/bazel-remote) proxy to store cache in Azure storage blobs and only 2Gb on disk(agents have 10gb available).
Linux is the fastest OS to build now with the most tests: https://github.com/microsoft/DeepGNN/actions/runs/5898231305

### DIFF
--- a/.github/actions/benchmarks/action.yml
+++ b/.github/actions/benchmarks/action.yml
@@ -4,8 +4,8 @@ runs:
   using: "composite"
   steps:
     - run: |
-        bazel run -c opt src/cc/lib/benchmark:grpc_benchmark --config=linux
-        bazel run -c opt src/cc/lib/benchmark:sampler_benchmark --config=linux
-        bazel run -c opt src/cc/lib/benchmark:neighbor_sampler_benchmark --config=linux
+        bazel run -c opt src/cc/lib/benchmark:grpc_benchmark --config=linux --remote_cache=http://localhost:8080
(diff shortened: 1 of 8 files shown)


.github/actions/benchmarks/action.yml
```diff
@@ -4,8 +4,8 @@ runs:
   using: "composite"
   steps:
     - run: |
-        bazel run -c opt src/cc/lib/benchmark:grpc_benchmark --config=linux
-        bazel run -c opt src/cc/lib/benchmark:sampler_benchmark --config=linux
-        bazel run -c opt src/cc/lib/benchmark:neighbor_sampler_benchmark --config=linux
+        bazel run -c opt src/cc/lib/benchmark:grpc_benchmark --config=linux --remote_cache=http://localhost:8080
```
coledie, line 7: Are these supposed to be localhost, it looks different from the other ones?

coledie (approved):

.github/actions/benchmarks/action.yml
mortid0: Yes, others also have a proxy set on local host. We don't use an env variable here, because this action only runs on linux and argument is explicit.

PR #225, opened 2023-09-25
[Remote multigraph 1/5] Add new keys to meta.json.

coledie:

New Behavior
----------------
Add new keys necessary for multigraph support to load and manage graphs. Make them optional since normal graphs do not need.

### DIFF
--- a/src/cc/lib/graph/metadata.h
+++ b/src/cc/lib/graph/metadata.h
@@ -19,6 +19,10 @@ struct Metadata
     explicit Metadata(std::filesystem::path path, std::string config_path = "");
     void Write(std::filesystem::path path) const;
 
+    // Optional graph information.
+    std::string m_name = "";
+    std::string m_data_path = "";
+
     // Graph information.
     size_t m_version;
     size_t m_node_count;
--- a/src/cc/lib/graph/metadata.h
+++ b/src/cc/lib/graph/metadata.h
@@ -21,7 +21,7 @@ struct Metadata
 
     // Optional graph information.
     std::string m_name = "";
-    std::string m_data_path = "";
+    std::vector<std::string> m_data_path = std::vector<std::string>();
 
     // Graph information.
     size_t m_version;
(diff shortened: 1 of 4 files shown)


coledie:
1. Do we need to update parameters of MultiWorkersConverter or anything to accept these parameters or should we expect users to add them manually?
2. Any unit tests needed for this? It seems meta.json is not directly tested anywhere?

src/cc/lib/graph/metadata.h
```diff
@@ -19,6 +19,10 @@ struct Metadata
     explicit Metadata(std::filesystem::path path, std::string config_path = "");
     void Write(std::filesystem::path path) const;
 
+    // Optional graph information.
+    std::string m_name = "";
+    std::string m_data_path = "";
```
mortid0, line 24: Server already accepts [a data location to load graph](https://github.com/microsoft/DeepGNN/blob/13f0b009cc78f147c8cb623e8205317e48aa7fe9/src/python/deepgnn/graph_engine/snark/server.py#L49) from a different location other than meta file. How conflicts are going to be resolved?

coledie: There will not be conflicts, we will only use meta.data_path if server is in active_monitor mode and it detects a meta.json in the monitored folder. But I will make it so the meta file supports the list of partition data dirs instead of just a single one though

mortid0: Please, try to avoid multiple config entries for data sources. I.e. if server is in active_monitor state and was started with data locations specified from command line, then we are going to monitor a new location specified in data_path?

mortid0:
> * Do we need to update parameters of MultiWorkersConverter or anything to accept these parameters or should we expect users to add them manually?
> * Any unit tests needed for this? It seems meta.json is not directly tested anywhere?

You probably don't need to change meta.json at all. Graph serving is independent from conversion and it should have its own config.

PR #229, opened 2023-09-27
Add PyG remote backend example.

coledie:

New Behavior
----------------
Add example with PyG remote backend example.


.github/workflows/ci.yml
```diff
@@ -127,3 +127,5 @@ jobs:
         run: cd ./tensorflow/han/ && sudo chmod +x *.sh && bash ./run.sh CPU cleanup
       - name: Run gat examples
         run: cd ./tensorflow/gat/ && sudo chmod +x *.sh && bash ./run.sh CPU cleanup
+      - name: Run PyG example
+        run: cd ./pytorch/pyg/ && sudo chmod +x *.py && python main.py
```
mortid0, line 131: all torch examples are run with tests now:
https://github.com/microsoft/DeepGNN/blob/a57ca28fe81d68196af57a4f4a9559f453febd44/.github/actions/test/action.yml#L34

examples/pytorch/pyg_interface.py
```diff
@@ -0,0 +1,203 @@
+"""PyGeo interface training example."""
```
mortid0, line 1: forgot copyright

examples/pytorch/pyg_interface.py
```diff
@@ -0,0 +1,203 @@
...
+    for data in tqdm.tqdm(loader):
+        data = data.to(device)
+        data = data[("0", "0", "0")]
+        optimizer.zero_grad()
+
+        h = model(data.x, data.edge_index)
+
+        h_src = h[data.edge_label_index[0]]
+        h_dst = h[data.edge_label_index[1]]
+        link_pred = (h_src * h_dst).sum(dim=-1)  # Inner product.
```
mortid0, line 120: Please, avoid [copy pasting code](https://github.com/pyg-team/pytorch_geometric/blob/02cc18d9d6841ecda4eb61eebb48b86f1aaae477/examples/graph_sage_unsup_ppi.py#L52) from other repos even with source comments

coledie: The problem is this interface is meant to be a remote backend for PyG so I need to use their model and dataloader and just copy the glue code from an example since it is not importable. What should I do instead?

mortid0: Write train loop on your own that is similar to theirs.

mortid0 (approved):

PR #233, opened 2023-10-04
Deprecate deepgnn-torch and deepgnn-tf packages

mortid0:

Previous Behavior
----------------
deepgnn-pytorch and deepgnn-tf packages were created with setup.py

New Behavior
----------------
1. Only deepgnn-ge package will be published
2. TF bazel targets are marked as deprecated to have a warning we are about to delete them from source code.
3. Don't run TF targets in CI, because of 2.


.github/ISSUE_TEMPLATE
```diff
@@ -4,8 +4,6 @@ Environment
 -----------
 * Python version: (python -V)
 * deepgnn-ge Version: (python -m pip show deepgnn-ge)
```
coledie, line 6: Just deepgnn now?

src/python/setup.py
```diff
@@ -211,9 +129,5 @@ def deepgnn_pytorch(version: str):
     target = target.lower()
     if target == "deepgnn-ge":
```
coledie, line 130: Remove target parameter completely, since we only have one target.

coledie (changes requested):

.github/ISSUE_TEMPLATE
mortid0: Just deepgnn means we need to release a new package. It might be useful later, but in this PR we want only to stop publishing -tf/-torch packages.

coledie: I see what you are saying but if we do them one at a time, users will have to update their pipelines and local installs twice deepgnn-torch -> deepgnn-ge -> deepgnn

mortid0: Only if we rename package to deepgnn as you are suggesting. We are not changing existing deepgnn-torch packages, but stop publishing new versions.

coledie (approved):

PR #247, opened 2023-12-20
Update hadoop version.

coledie:

New Behavior
----------------
Update HDFS version since old one has been deleted from official source.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -160,9 +160,9 @@ http_archive(
 http_archive(
     name = "hadoop",
     build_file = "//config:hadoop.BUILD",
-    sha256 = "ad770ae3293c8141cc074df4b623e40d79782d952507f511ef0a6b0fa3097bac",
-    strip_prefix = "hadoop-3.3.1",
-    urls = ["https://dlcdn.apache.org/hadoop/common/hadoop-3.3.1/hadoop-3.3.1.tar.gz"],
+    sha256 = "f5195059c0d4102adaa7fff17f7b2a85df906bcb6e19948716319f9978641a04",
+    strip_prefix = "hadoop-3.3.6",
+    urls = ["https://dlcdn.apache.org/hadoop/common/hadoop-3.3.6/hadoop-3.3.6.tar.gz"],
 )
 
 http_archive(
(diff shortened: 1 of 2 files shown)


mortid0 (approved):
Thanks for a quick fix!

.github/actions/wheel/action.yml
```diff
@@ -15,7 +15,10 @@ runs:
       run: |
         echo "BAZEL_CONFIG=manylinux" >> $GITHUB_ENV
         echo "PLAT_NAME=manylinux1_x86_64" >> $GITHUB_ENV
-        echo "REMOTE_CACHE=--remote_cache=http://localhost:8080" >> $GITHUB_ENV
+        if ping -c 1 http://localhost:8080 &> /dev/null
```
coledie, line 18: Cache breaks pretty often, we need to just check if its working at the start and decide to actually use it or not based on that.

coledie:
@mortid0 ready for review

.github/actions/buildcache/action.yml
```diff
@@ -13,3 +13,4 @@ runs:
         go install github.com/buchgr/bazel-remote/v2@v2.4.1
         bazel-remote --azblob.tenant_id=bazelcache --azblob.storage_account=bazelcache --azblob.container_name=cache --azblob.auth_method=shared_key --dir=/tmp/bazelcache --max_size=2 --http_address=localhost:8080 &
         sleep 5 # Give the server time to start
+        echo $(curl ${{ inputs.cache_address }}/status)
```
mortid0, line 16: why do we need this? Should we just increase the sleep above?

coledie: No, if you actually ping the status - it works but it says 0 files loaded. Even if you check status after an hour at the start of benchmarks, its 0 files. Must be auth issue or something - either way, the cache has broken several times and it completely stops the CI from working so this catches that. A separate PR should be used to fix the cache connection.

mortid0: ah, we need to update `BAZEL_REMOTE_AZBLOB_SHARED_KEY` then.

mortid0 (approved):

PR #254, opened 2024-01-18
Fix edge feature fetching

mortid0:

Previous Behavior
----------------
If edges were not sorted, then we failed to fetch edge features in cases where binary search will not find a correct destination.

New Behavior
----------------
1. Sort edges by destination in a converter for non-temporal graphs.
2. Fallback to linear search, when binary search does not find anything.

### DIFF
--- a/src/cc/lib/benchmark/search_benchmark.cc
+++ b/src/cc/lib/benchmark/search_benchmark.cc
@@ -1,0 +1,68 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <algorithm>
+#include <iostream>
+#include <numeric>
+#include <random>
+#include <string>
+#include <vector>
+
+#include <benchmark/benchmark.h>
+
+#ifdef SNARK_PLATFORM_LINUX
+#include <mimalloc-override.h>
+#endif
+
+static void BM_LOWER_BOUND(benchmark::State &state)
+{
+    size_t max_count = state.range(0);
+    std::vector<int64_t> elements(max_count);
+    std::iota(elements.begin(), elements.end(), 0);
+    std::vector<int64_t> ids(elements);
+    int64_t seed = 42;
+    std::shuffle(ids.begin(), ids.end(), std::mt19937_64(seed));
+    size_t curr = 0;
+    for (auto _ : state)
+    {
+        ++curr;
+        if (curr == max_count)
+        {
+            curr = 0;
+        }
+        const auto res = std::lower_bound(elements.cbegin(), elements.cend(), ids[curr]);
+        if (*res != ids[curr])
+        {
+            throw std::runtime_error("not found" + std::to_string(ids[curr]) + " got " + std::to_string(*res));
+        }
+    }
+}
+
+static void BM_STD_FIND(benchmark::State &state)
+{
+    size_t max_count = state.range(0);
+    std::vector<int64_t> elements(max_count);
+    std::iota(elements.begin(), elements.end(), 0);
+    std::vector<int64_t> ids(elements);
+    int64_t seed = 42;
+    std::shuffle(ids.begin(), ids.end(), std::mt19937_64(seed));
+    size_t curr = 0;
+    for (auto _ : state)
+    {
+        ++curr;
+        if (curr == max_count)
+        {
+            curr = 0;
+        }
+        const auto res = std::find(elements.cbegin(), elements.cend(), ids[curr]);
+        if (*res != ids[curr])
+        {
+            throw std::runtime_error("not found" + std::to_string(ids[curr]) + " got " + std::to_string(*res));
+        }
+    }
+}
+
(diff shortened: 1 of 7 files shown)


src/cc/lib/benchmark/search_benchmark.cc
```diff
@@ -0,0 +1,68 @@
...
+
+#include <benchmark/benchmark.h>
+
+#ifdef SNARK_PLATFORM_LINUX
+#include <mimalloc-override.h>
+#endif
+
+static void BM_LOWER_BOUND(benchmark::State &state)
+{
+    size_t max_count = state.range(0);
```
coledie, line 19: Where does this use our code? Is it just benchmark on rng?

mortid0: It is how we picked this magic const:
image

coledie: I see, maybe a docstring would be helpful here or there for future matinence

src/cc/lib/graph/partition.cc
```diff
@@ -908,16 +908,38 @@ std::optional<size_t> Partition::EdgeFeatureOffset(uint64_t internal_src_node_id
...
+        if (it != lst && *it == input_edge_dst)
+        {
+            return {it - std::begin(m_edge_destination)};
+        }
+
         return std::nullopt;
     }
-    return {it - std::begin(m_edge_destination)};
+
+    auto it = std::lower_bound(fst, lst, input_edge_dst);
```
coledie, line 928: I don't see any validation that m_edge_destination is sorted per lower_bound requirements - at least not in the C++ code, I'm sure in python converter it is, but should we have such verification if algo depends on it?

src/cc/lib/graph/partition.cc
```diff
@@ -908,16 +908,38 @@ std::optional<size_t> Partition::EdgeFeatureOffset(uint64_t internal_src_node_id
...
+    if (it != lst && *it == input_edge_dst)
+    {
+        return {it - std::begin(m_edge_destination)};
+    }
+
+    // For temporal graphs we need to use fallback to linear search, since edges are stored based on timestamps first
+    // order.
+    it = std::find(fst, lst, input_edge_dst);
```
coledie, line 936: Can we at least filter input_edge_dst by active timestamps only so we reduce linear search time?

src/cc/lib/graph/partition.cc
mortid0: We don’t need this validation. Plain fallback to a linear search when element not found works.

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
```diff
@@ -1431,6 +1431,93 @@ def test_edge_index_inverted_types(graph_with_inverse_edge_type_order):
...
+
+
+@pytest.fixture(scope="module")
+def graph_with_inverse_edge_dst_order(request):
+    workdir = tempfile.TemporaryDirectory()
+    if request.param == JsonDecoder:
+        data_name = graph_with_inverse_edge_dst_order_json(workdir.name)
+    elif request.param == TsvDecoder:
+        data_name = graph_with_inverse_edge_dst_order_tsv(workdir.name)
+    else:
```
coledie, line 1481: Need for edge list as well, can just use json to edgelist converter already in file.

src/cc/lib/graph/partition.cc
mortid0: If you mean filter and create a new vector, then it is an allocation, which is slow. Without an extra variable it will be a very complex binary search.

mortid0: This approach doesn’t require binary data regeneration for non temporal graphs and is backward compatible.
We can write a complex binary search for temporal graphs with edge features when it becomes a bottleneck.

src/cc/lib/graph/partition.cc
```diff
@@ -908,16 +908,38 @@ std::optional<size_t> Partition::EdgeFeatureOffset(uint64_t internal_src_node_id
...
     auto fst = std::begin(m_edge_destination) + m_edge_type_offset[type_offset];
     auto lst = fst + tp_count;
-    auto it = std::lower_bound(fst, lst, input_edge_dst);
-    if (it == lst)
+
+    // Fast path: linear search outperforms binary search for small number of elements.
+    if (tp_count <= 16)
     {
-        // Edge was not found in this partition.
+        auto it = std::find(fst, lst, input_edge_dst);
```
coledie, line 919: This is repeated code with temporal case on line 936, just invert if then merge

src/python/deepgnn/graph_engine/snark/tests/convert_test.py
mortid0: EdgeList produces edges, I don't think we can sort it and test..

coledie (approved):

PR #323, opened 2024-12-24
[FEATURE] Add skip temporal loading flag

Tomcli:
We propose adding a flag to the DeepGNN Snark Server that allows us to opt out of loading watermark without modifying the meta.json. Our dataset watermarks constitute more than 30% of our data which could waste a lot of GE memory resources. For models that don't require the watermark, we want to configure the GE to avoid loading the unused watermark. This will enable faster loading times and save resources for hosting GEs.

Previous Behavior
----------------
GE have to load all the watermarks unless manually change the meta.json that's shared with other developers/teams.

New Behavior
----------------
Users can pick whether to load features or watermark based on their model training settings.

### DIFF
--- a/src/cc/lib/graph/metadata.cc
+++ b/src/cc/lib/graph/metadata.cc
@@ -62,7 +62,6 @@ Metadata::Metadata(std::filesystem::path path, std::string config_path, bool ski
     // Default meta.json values
     m_node_feature_count = 0;
     m_edge_feature_count = 0;
-    m_watermark = -1;
     m_node_count = meta["node_count"];
     m_edge_count = meta["edge_count"];
     m_node_type_count = meta["node_type_count"];
--- a/src/cc/lib/graph/metadata.cc
+++ b/src/cc/lib/graph/metadata.cc
@@ -17,7 +17,7 @@ namespace snark
 {
 
 Metadata::Metadata(std::filesystem::path path, std::string config_path, bool skip_feature_loading,
-                   std::shared_ptr<Logger> logger)
+                   bool skip_watermark_loading, std::shared_ptr<Logger> logger)
     : m_version(MINIMUM_SUPPORTED_VERSION), m_path(path.string()), m_config_path(config_path), m_watermark(-1)
 {
 
@@ -72,6 +72,12 @@ Metadata::Metadata(std::filesystem::path path, std::string config_path, bool ski
     {
         m_node_feature_count = 0;
         m_edge_feature_count = 0;
+    }
+    // Skip watermark loading if requested. This is useful when the watermark is not needed for the training/inference
+    // job.
+    if (skip_watermark_loading == true)
+    {
+        m_watermark = -1;
     }
 
     m_partition_node_weights =
--- a/src/cc/lib/graph/metadata.cc
+++ b/src/cc/lib/graph/metadata.cc
@@ -18,8 +18,7 @@ namespace snark
 
 Metadata::Metadata(std::filesystem::path path, std::string config_path, bool skip_feature_loading,
                    bool skip_temporal_loading, std::shared_ptr<Logger> logger)
-    : m_version(MINIMUM_SUPPORTED_VERSION), m_path(path.string()), m_config_path(config_path), m_watermark(-1),
-      m_node_feature_count(0), m_edge_feature_count(0)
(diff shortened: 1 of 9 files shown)


Tomcli:
The CI seems to be broken in the recent master branch which is irrelevant to this PR. The CI test data cannot be loaded for the hdfs test.

No rush, this PR can be reviewed after the holidays.

src/cc/lib/graph/metadata.cc
```diff
@@ -73,6 +73,12 @@ Metadata::Metadata(std::filesystem::path path, std::string config_path, bool ski
         m_node_feature_count = 0;
         m_edge_feature_count = 0;
     }
+    // Skip watermark loading if requested. This is useful when the watermark is not needed for the training/inference
+    // job.
+    if (skip_watermark_loading == true)
```
coledie, line 78: Watermark is already -1, we should skip trying to read it from the meta instead of reading it from meta then setting it to -1 again.

src/cc/lib/graph/metadata.cc
```diff
@@ -17,7 +17,7 @@ namespace snark
 {
 
 Metadata::Metadata(std::filesystem::path path, std::string config_path, bool skip_feature_loading,
-                   std::shared_ptr<Logger> logger)
+                   bool skip_watermark_loading, std::shared_ptr<Logger> logger)
```
coledie, line 20: It seems like the goal is to take some data that is generated as temporal and turn it to a regular non-temporal dataset? Can we adjust the parameter name accordinglty so its more clear, probably no one will know what a watermark is

coledie (changes requested):
comments

coledie:
> The CI seems to be broken in the recent master branch which is irrelevant to this PR. The CI test data cannot be loaded for the hdfs test.
> 
> No rush, this PR can be reviewed after the holidays.

This is in progress, we can still merge this I can manually check tests.

src/cc/lib/graph/metadata.cc
Tomcli: Thanks, I updated the code to only load the meta if skip feature/temporal loading is not set to true.

src/cc/lib/graph/metadata.cc
Tomcli: > It seems like the goal is to take some data that is generated as temporal and turn it to a regular non-temporal dataset?
Yes, our data pipelines currently generate everything with temporal edges/features. Our temporal edges are consuming around 40% of our GE memory, which most of our models do not utilize. Therefore, we want to have the option to skip loading the temporal dataset for models that do not require it.

I renamed the parameter to skip_temporal_loading.

Tomcli:
Thank you @coledie for the review.

coledie (changes requested):
Once these comments are commited, I will test and merge. I have the PR to fix the dataset fetching issues started. Do you want me to cut a release for this change only or will there be more soon?

src/cc/lib/graph/metadata.cc
```diff
@@ -59,19 +59,27 @@ Metadata::Metadata(std::filesystem::path path, std::string config_path, bool ski
                           m_version, MINIMUM_SUPPORTED_VERSION);
     }
 
+    // Default meta.json values
+    m_node_feature_count = 0;
+    m_edge_feature_count = 0;
+    m_watermark = -1;
```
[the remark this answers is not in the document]
Tomcli: Thanks I removed this line

src/cc/lib/py_server.cc
```diff
@@ -32,10 +32,11 @@ std::string safe_convert(const char *buffer)
 int32_t StartServer(PyServer *graph, const char *meta_location, size_t count, uint32_t *partition_indices,
                     const char **partition_locations, const char *host_name, const char *ssl_key, const char *ssl_cert,
                     const char *ssl_root, const PyPartitionStorageType storage_type_, const char *config_path,
-                    bool skip_feature_loading)
```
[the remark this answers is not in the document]
Tomcli: Replicated this unit test with the skip temporal metadata flag.

Tomcli:
> Once these comments are commited, I will test and merge. I have the PR to fix the dataset fetching issues started. Do you want me to cut a release for this change only or will there be more soon?

Thank you @coledie. This is the only change we want for the next release. Our objective is to provide ways to minimize resource consumption for GEs so that we can scale our graph without incurring additional costs.

